PR #77, opened 2014-11-17
Charm storage metadata

axw:
We introduce a new "storage" attribute to charm metadata, which describes required and optional storage for a charm. Each may be raw block devices, or file systems. The charm declares the name, type and some restrictions on the storage, and the administrator specifies how to fulfil that storage when deploying the charm.

The metadata format is based on the storage phase 1 scope document:
https://docs.google.com/a/canonical.com/document/d/1-9ZPfdgpkj2R9mBG_tlSclGGyK3tRpMf2L4C37mzYD8/edit#bookmark=id.6abw2ee0ebgm

which is mostly the same as described in the spec: https://docs.google.com/a/canonical.com/document/d/1OhaLiHMoGNFEmDJTiNGMluIlkFtzcYjezC8Yq4nAX3Q/edit

Note that the spec was never signed off, and the scope document is yet to be reviewed. This metadata format is subject to change, and should not be documented until it is finalised. In the short term, this unblocks further work on storage in Juju (which should be uncontroversial).

There are also two new hook kinds defined for notifying units of storage attachment and detachment.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -52,33 +52,33 @@ type Storage struct {
 	// Name is the name of the storage requirement.
 	//
 	// Name has no default, and must be specified.
-	Name string
+	Name string `bson:"name"`
 
 	// Type is the storage type: filesystem or block-device.
 	//
 	// Type has no default, and must be specified.
-	Type StorageType
+	Type StorageType `bson:"type"`
 
 	// Shared indicates that the storage is shared between all units of
 	// a service deployed from the charm. It is an error to attempt to
 	// assign non-shareable storage to a "shared" storage requirement.
 	//
 	// Shared defaults to false.
-	Shared bool
+	Shared bool `bson:"shared"`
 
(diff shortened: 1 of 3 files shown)


axw:
PTAL. I've updated again to match recent changes to the spec.
- No more filesystem (i.e. filesystem type preference)
- No more persistent (this is configured by the user, on the storage pool)
- Change count to multiple.range
- Add description field

I have not yet added minimum size or properties: these can be added later.

meta.go
```diff
@@ -37,6 +39,89 @@ const (
...
+// StorageType defines a storage type.
+type StorageType string
+
+const (
+	StorageBlock      StorageType = "block"
+	StorageFilesystem StorageType = "filesystem"
+)
+
+// Storage represents a charm's storage requirement.
+type Storage struct {
```
fwereade: yuck, but, ok :/

fwereade:
LGTM

PR #88, opened 2015-03-03
Add support for parsing more charm storage metadata

wallyworld:
minimum-size and properties are now supported in charm storage metadata

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/juju/schema"
+	"github.com/juju/utils"
 	"gopkg.in/yaml.v1"
 
 	"gopkg.in/juju/charm.v4/hooks"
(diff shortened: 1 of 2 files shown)


meta.go
```diff
@@ -92,12 +93,30 @@ type Storage struct {
...
 	// Location has no default, and is optional.
 	Location string `bson:"location,omitempty"`
+
+	// Properties allow the charm author to characterise the relative storage
+	// performance requirements and sensitivities for each store.
+	// eg “transient” is used to indicate that non persistent storage is acceptable,
+	// such as tmpfs or ephemeral instance disks.
+	//
+	// Properties has no default, and is optional.
+	Properties []string `bson:properties,omitempty`
```
axw: Given that we won't be honouring properties yet, and it's not clear in the spec whether properties are requirements or hints (I opened a question, never got answered), I'd really rather if this got left out for now.

meta_test.go
```diff
@@ -675,6 +675,18 @@ storage:
...
+	}, {
+		desc: "minimum size must parse correctly",
+		yaml: "  type: block\n  minimum-size: foo",
+		err:  `metadata: expected a non-negative number, got "foo"`,
+	}, {
+		desc: "minimum size must have valid suffix",
+		yaml: "  type: block\n  minimum-size: 10Q",
+		err:  `metadata: invalid multiplier suffix "Q", expected one of MGTPEZY`,
+	}, {
+		desc: "properties must contain valid valies must have valid suffix",
```
axw: wat

axw:
LGTM, but I think it's premature to add `properties`. It is not well defined, and we don't know how it'll be used yet.

meta.go
wallyworld: As discussed live, the properties are used as hints and jam is ok with implementing these now.

PR #90, opened 2015-03-12
Reorganize legacy charmrepo code.

frankban:
This branch only includes moved code.
Move current charmstore impl to legacy.go, and the local repository impl to local.go.
New charmstore code will be implemented in a future charmstore.go file.


charmrepo/legacy.go
```diff
@@ -0,0 +1,349 @@
+// Copyright 2012, 2013 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package charmrepo
```
rogpeppe: // Package charmrepo implements access to charm repositories.

?

mhilton:
:+1: LGTM

rogpeppe:
LGTM with one trivial suggestion

charmrepo/legacy.go
frankban: Done.

PR #94, opened 2015-03-17
Implement the repo.Latest method of the new charm store.

frankban:


charmrepo/charmstore.go
```diff
@@ -71,7 +74,58 @@ func (s *CharmStore) Get(curl *charm.URL) (charm.Charm, error) {
...
+
+	// Prepare the request to the charm store.
+	urls := make([]string, len(curls))
+	ids := make([]string, len(curls))
+	for i, curl := range curls {
+		url := curl.WithRevision(-1).String()
+		urls[i] = url
+		ids[i] = "id=" + url
+	}
+	path := "/meta/any?include=id-revision&include=hash256&" + strings.Join(ids, "&")
```
mhilton: Perhaps you could build this with a URL e.g.
```
var u url.URL
u.Path = "/meta/any"
var v url.Values
v.Add("include", "id-revision")
v.Add("include", "hash256")
for i, curl := range curls {
        url := curl.WithRevision(-1).String()
        urls[i] = url
        v.Add("id", url)
}
u.RawQuery = v.String()
path = u.String()
```
?

mhilton:
:+1: LGTM with a suggestion.

charmrepo/charmstore.go
```diff
@@ -71,7 +74,58 @@ func (s *CharmStore) Get(curl *charm.URL) (charm.Charm, error) {
...
+
+	// Execute the request and retrieve results.
+	var results map[string]struct {
```
rogpeppe: I think you could avoid some work here by declaring
the retrieved fields explicitly:

    var results map[string] struct {
      Meta struct {
           IdRevision params.IdRevisionResponse `json:"id-revision"`
           Hash256 params.HashResponse `json:"hash256"`
      }
    }

Even better would be to add a bulk meta method as a companion
to csclient.Client.Meta.

charmrepo/charmstore.go
rogpeppe: I had been thinking something similar.

rogpeppe:
LGTM with one suggestion, thanks!

charmrepo/charmstore.go
frankban: Nice suggestion! Done.
csclient.BulkMeta is also a good idea, I'll investigate that later.

charmrepo/charmstore.go
frankban: Nice, Done.

frankban:
Thanks for the reviews!

PR #95, opened 2015-03-18
Implement charmstore.Resolve.

frankban:
It can be used to resolve charm series.

### DIFF
--- a/charmrepo/charmstore_test.go
+++ b/charmrepo/charmstore_test.go
@@ -190,6 +190,59 @@ func (s *charmStoreRepoSuite) TestLatest(c *gc.C) {
 	}
 }
 
+func (s *charmStoreRepoSuite) TestResolve(c *gc.C) {
+	// Add some charms to the charm store.
+	s.addCharm(c, "~who/trusty/mysql", "mysql")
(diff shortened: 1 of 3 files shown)


mhilton:
:+1: LGTM

charmrepo/charmstore_test.go
```diff
@@ -190,6 +190,59 @@ func (s *charmStoreRepoSuite) TestLatest(c *gc.C) {
...
+		url: "cs:~who/precise/wordpress",
+	}, {
+		id:  "~who/wordpress-1",
+		url: "cs:~who/precise/wordpress-1",
+	}, {
+		id:  "~dalek/riak",
+		url: "cs:~dalek/utopic/riak",
+	}, {
+		id:  "~dalek/utopic/riak-2",
+		url: "cs:~dalek/utopic/riak-2",
```
urosj: Not that it matters, but should this be -3 as -2 holds "wordpress" content?

urosj:
:+1: Ty.

charmrepo/charmstore_test.go
frankban: It holds wordpress content but that content is still associated with the riak reference.

PR #96, opened 2015-03-19
Add series to machinespec

makyo:
This branch adds an optional series attribute to MachineSpec, which allows a user to specify a machine of a particular series for creation.  This also includes a drive-by fix to ActionSpec.InsertDefaults so that return types match.


mhilton:
:+1: LGTM

frankban:
:+1: I wonder if we should also validate that, if a service is hulk smashed in a machine with ad defined series, the service charm series must match. Not for this branch.

PR #97, opened 2015-03-20
Implement CharmStore.Get.

frankban:
CharmStore.Get is used to retrieve a charm
from the charm store.

### DIFF
--- a/charmrepo/charmstore.go
+++ b/charmrepo/charmstore.go
@@ -71,16 +71,15 @@ func NewCharmStore(p NewCharmStoreParams) (Interface, error) {
 
 // Get implements Interface.Get.
 func (s *CharmStore) Get(curl *charm.URL) (charm.Charm, error) {
-	// Disable bundle deployment.
 	if curl.Series == "bundle" {
 		return nil, errgo.Newf("expected a charm URL, got bundle URL %q", curl)
 	}
 
 	// Prepare the cache directory and retrieve the charm.
-	if err := os.MkdirAll(s.cacheDir, os.FileMode(0755)); err != nil {
+	if err := os.MkdirAll(s.cacheDir, 0755); err != nil {
 		return nil, errgo.Notef(err, "cannot create the cache directory")
 	}
-	r, id, expectHash, _, err := s.client.GetArchive(curl.Reference())
+	r, id, expectHash, expectSize, err := s.client.GetArchive(curl.Reference())
 	if err != nil {
 		return nil, errgo.Notef(err, "cannot retrieve charm")
 	}
@@ -88,7 +87,7 @@ func (s *CharmStore) Get(curl *charm.URL) (charm.Charm, error) {
 
 	// Check if the archive already exists in the cache.
 	path := filepath.Join(s.cacheDir, charm.Quote(id.String())+".charm")
-	if verifyHash384(path, expectHash) == nil {
+	if verifyHash384AndSize(path, expectHash, expectSize) == nil {
 		return charm.ReadCharmArchive(path)
 	}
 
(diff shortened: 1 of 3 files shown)


charmrepo/charmstore.go
```diff
@@ -63,11 +69,65 @@ func NewCharmStore(p NewCharmStoreParams) (Interface, error) {
 	}, nil
 }
 
-var notImplemented = errgo.New("not implemented")
-
 // Get implements Interface.Get.
 func (s *CharmStore) Get(curl *charm.URL) (charm.Charm, error) {
-	return nil, notImplemented
+	// Disable bundle deployment.
```
rogpeppe: This comment seems somewhat redundant to me

charmrepo/charmstore.go
```diff
@@ -63,11 +69,65 @@ func NewCharmStore(p NewCharmStoreParams) (Interface, error) {
...
+	if curl.Series == "bundle" {
+		return nil, errgo.Newf("expected a charm URL, got bundle URL %q", curl)
+	}
+
+	// Prepare the cache directory and retrieve the charm.
+	if err := os.MkdirAll(s.cacheDir, os.FileMode(0755)); err != nil {
```
rogpeppe: s/os.FileMode//

charmrepo/charmstore.go
```diff
@@ -63,11 +69,65 @@ func NewCharmStore(p NewCharmStoreParams) (Interface, error) {
...
+	}
+	r, id, expectHash, _, err := s.client.GetArchive(curl.Reference())
+	if err != nil {
+		return nil, errgo.Notef(err, "cannot retrieve charm")
+	}
+	defer r.Close()
+
+	// Check if the archive already exists in the cache.
+	path := filepath.Join(s.cacheDir, charm.Quote(id.String())+".charm")
+	if verifyHash384(path, expectHash) == nil {
```
rogpeppe: I wonder if we should be using sha256 for this so that we don't blow everyone's cache.

mhilton:
:+1: LGTM thanks

charmrepo/charmstore.go
```diff
@@ -63,11 +69,65 @@ func NewCharmStore(p NewCharmStoreParams) (Interface, error) {
...
+	// Disable bundle deployment.
+	if curl.Series == "bundle" {
+		return nil, errgo.Newf("expected a charm URL, got bundle URL %q", curl)
+	}
+
+	// Prepare the cache directory and retrieve the charm.
+	if err := os.MkdirAll(s.cacheDir, os.FileMode(0755)); err != nil {
+		return nil, errgo.Notef(err, "cannot create the cache directory")
+	}
+	r, id, expectHash, _, err := s.client.GetArchive(curl.Reference())
```
rogpeppe: I think you should use the size here and check that the number of bytes returned by io.Copy is as expected. An error message that says there's a size mismatch, giving expected and actual sizes, is going to be much more accessible than a hash mismatch message (which of course needs to be done too).

charmrepo/charmstore.go
rogpeppe: perhaps log the error message on failure?

rogpeppe:
LGTM with a couple of trivial suggestions, thanks!

charmrepo/charmstore.go
frankban: Removed.

charmrepo/charmstore.go
frankban: Done.

charmrepo/charmstore.go
frankban: Done.

charmrepo/charmstore.go
frankban: Done.

PR #99, opened 2015-03-23
Restore infer repository for legacy repo.

frankban:
Also rename Store to LegacyStore.

### DIFF
--- a/charmrepo/legacy_test.go
+++ b/charmrepo/legacy_test.go
@@ -371,6 +371,37 @@ func (s *legacyCharmStoreSuite) TestCharmURL(c *gc.C) {
 	}
 }
 
+var legacyInferRepositoryTests = []struct {
+	url  string
+	path string
+}{
+	{"cs:precise/wordpress", ""},
+	{"local:oneiric/wordpress", "/some/path"},
+}
+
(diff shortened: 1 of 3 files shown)


makyo:
:+1:

charmrepo/legacy_test.go
```diff
@@ -371,6 +371,37 @@ func (s *legacyCharmStoreSuite) TestCharmURL(c *gc.C) {
...
+	}
+	ref, err := charm.ParseReference("local:whatever")
+	c.Assert(err, gc.IsNil)
+	_, err = charmrepo.LegacyInferRepository(ref, "")
+	c.Assert(err, gc.ErrorMatches, "path to local repository not specified")
+	ref.Schema = "foo"
+	_, err = charmrepo.LegacyInferRepository(ref, "")
+	c.Assert(err, gc.ErrorMatches, "unknown schema for charm reference.*")
+}
+
```
urosj: What if "cs:something" is passed the path string, is the paths skipped and only CS part taken for inferring repository?

urosj:
:+1: Small comment, but the case is probably filtered on the client level anyway.

PR #100, opened 2015-03-24
Fix mock store test mode handling.

frankban:
The mock store defines SetTestMode while the real legacy charm store uses WithTestMode.
This historically meant false positives on juju-core tests, with test mode not really working.


bac:
:+1:

jrwren:
:+1:

frankban:
Thanks for the reviews!

PR #106, opened 2015-03-27
Use the charmstore own testing server in tests.

frankban:
This must land after https://github.com/juju/charmstore/pull/330

### DIFF
--- a/charmrepo/charmstore_test.go
+++ b/charmrepo/charmstore_test.go
@@ -22,7 +22,7 @@ import (
 	"gopkg.in/juju/charmstore.v4"
 	"gopkg.in/juju/charmstore.v4/csclient"
 	"gopkg.in/juju/charmstore.v4/params"
-	"gopkg.in/mgo.v2"
+	charmstoretesting "gopkg.in/juju/charmstore.v4/testing"
 
 	"gopkg.in/juju/charm.v5-unstable"
 	"gopkg.in/juju/charm.v5-unstable/charmrepo"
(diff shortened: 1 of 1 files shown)


jrwren:
:+1:

charmrepo/charmstore_test.go
```diff
@@ -94,18 +94,55 @@ type charmStoreRepoSuite struct {
...
+
+	getDownloads := func() int {
+		var result []params.Statistic
+		err := client.Get(path, &result)
+		c.Assert(err, jc.ErrorIsNil)
+		return int(result[0].Count)
+	}
+
+	for retry := 0; retry < 10; retry++ {
+		time.Sleep(100 * time.Millisecond)
```
hatched: Is waiting the best way to do this? Should we not instead wait on a trigger of some sort?

hatched:
Thanks! I have just one small question if there is no better way then :+1:

charmrepo/charmstore_test.go
frankban: Yes, we are on the client side, and so we don't have events from the server.

frankban:
Thanks for the reviews!

charmrepo/charmstore_test.go
```diff
@@ -77,14 +77,14 @@ func (s *charmStoreBaseSuite) TearDownTest(c *gc.C) {
...
-		User:     serverParams.AuthUsername,
-		Password: serverParams.AuthPassword,
-	})
-	ch := TestCharms.CharmDir(name)
-	id, err := client.UploadCharm(charm.MustParseReference(url), ch)
-	c.Assert(err, jc.ErrorIsNil)
+	id := charm.MustParseReference(url)
+	promulgated := false
+	if id.User == "" {
+		id.User = "who"
```
urosj: Why do you set a user to "who"?

PR #114, opened 2015-04-08
Update charmstore repo tests to use the new charmstoretesting package.

frankban:


frankban:
Self review: trivial.

PR #120, opened 2015-04-14
Close file before renaming

gabriel-samfira:
We need to close the file before we move it on windows.


charmrepo/charmstore.go
```diff
@@ -102,7 +102,7 @@ func (s *CharmStore) Get(curl *charm.URL) (charm.Charm, error) {
 	if err != nil {
 		return nil, errgo.Notef(err, "cannot make temporary file")
 	}
-	defer f.Close()
```
axw: Don't remove this, because if we error early then the file isn't closed. It's not harmful to call Close twice.

gabriel-samfira:
Done. Thanks.

natefinch:
LGTM

PR #123, opened 2015-04-28
Allow NewCharmStore to be patched for testing

wallyworld:
Introduce a var for NewCharmStore to allow tests to patch a custom version of this function.


axw:
NOT LGTM
just use the original var expose/patch in your juju/juju branch, then we can fix it later without dirtying this API

PR #125, opened 2015-04-29
hooks: s/storage-detached/storage-detaching/

axw:
Rename storage-detached to storage-detaching. The hook will be run before storage is detached, to give charms an opportunity to cleanly release.


wallyworld:
LGTM

rogpeppe:
For the record, this is an API-breaking change and as such should not have landed
in charm.v5.

PR #137, opened 2015-06-10
Adjust workload process metadata to changes in the spec.

ericsnowcurrently:


process.go
```diff
@@ -182,28 +212,39 @@ func checkProcesses(procs map[string]Process) error {
...
+		"env":          schema.Omit,
 	},
 )
 
-type processTypeOptionChecker struct{}
+type forcedStringChecker struct{}
 
-func (c processTypeOptionChecker) Coerce(v interface{}, path []string) (interface{}, error) {
-	return fmt.Sprintf("%v", v), nil
+func (c forcedStringChecker) Coerce(v interface{}, path []string) (interface{}, error) {
```
kat-co: As a drive-by, please add a comment to this public method.

process.go
```diff
@@ -219,9 +260,17 @@ func (c processPortsChecker) Coerce(v interface{}, path []string) (interface{},
 		return nil, fmt.Errorf("%s: invalid value %q", strings.Join(path[1:], ""), item)
 	}
 
-	portA, err := strconv.Atoi(parts[0])
-	if err != nil {
-		return nil, fmt.Errorf("%s: expected int got %q", strings.Join(path[1:], ""), parts[0])
+	portA := 0
+	external := parts[0]
+	if !strings.HasPrefix(external, "<") || !strings.HasSuffix(external, ">") {
```
kat-co: It's not immediately evident what this chunk of code is trying to accomplish. Please add a few pointer comments.

process.go
```diff
@@ -58,12 +35,221 @@ type Process struct {
...
+		volumes = append(volumes, volume.Copy())
+	}
+	copied.Volumes = volumes
+
+	return copied
+}
+
+// ProcessFieldValue describes a requested change to a Process.
+type ProcessFieldValue struct {
+	Field    string
```
kat-co: These public fields need to be documented.

process.go
```diff
@@ -58,12 +35,221 @@ type Process struct {
...
+		p.EnvVars[value.Subfield] = value.Value
+	default:
+		return fmt.Errorf("unrecognized field %q", value.Field)
+	}
+	return nil
+}
+
+// Extend updates the Process with the provided value. If the
+// identified field is already set then Extend fails.
+func (p *Process) Extend(value ProcessFieldValue) error {
```
kat-co: This looks like the inverse of "Override(...)". Maybe extract the logic out into a private method, add a boolean (example name ``failIfExists``) parameter, and then change your checks to be:
```go
if p.Description != "" && failIfExists {
    return fmt.Errorf(`"description" already set`) // Tweak the error message to fit the failIfExists case
```

process.go
```diff
@@ -88,37 +295,152 @@ func (p Process) Validate() error {
...
+}
+
+// ProcessVolume is storage volume information for a workload process.
+type ProcessVolume struct {
+	// ExternalMount is the path on the host.
+	ExternalMount string
+	// InternalMount is the path on the process.
+	InternalMount string
+	// Mode is the "ro" OR "rw"
+	Mode string
```
kat-co: If it can only be two things, should we have a typedef?

ericsnowcurrently: I don't think so.  Since Go doesn't check the values, a typedef doesn't help much, particularly since it's a string.  In fact it kind of makes the code messier and harder to follow.  We have all the validation we need for the two values so I think it's fine as-is.  Given that we rely pretty heavily on serialization and parsing for the metadata, let's just keep this as strings.

process.go
ericsnowcurrently: As a whole there are subtle differences which mean it's not quite the inverse.  I actually started off trying to do what you are talking about and it ended up messier and harder to follow.  This way is clear and explicit.  If it turns out to be a problem (which I doubt) then we can refactor these pieces later.

ericsnowcurrently:
I believe we've addressed all your comments.  If you feel the responses are satisfactory then are we good to go?

kat-co:
LGTM

PR #140, opened 2015-06-24
new comment for TypeOptions

natefinch:
for me, at least, this makes it a lot more obvious what the TypeOptions field is for.


mattyw:
LGTM

PR #141, opened 2015-07-16
Support for builtin metrics

alesstimec:


mattyw:
LGTM

tasdomas:
LGTM

metrics.go
```diff
@@ -16,6 +17,8 @@ import (
 type MetricType string
 
 const (
+	builtinMetricsPrefix = "juju"
```
mattyw: We should export this an use it in core

PR #142, opened 2015-07-20
metrics: export a function to decide if a metric is builtin. 

mattyw:
Keeps the logic in one place

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -17,12 +17,17 @@ import (
 type MetricType string
 
 const (
-	builtinMetricsPrefix = "juju"
+	builtinMetricPrefix = "juju"
 
 	// Supported metric types.
 	MetricTypeGauge    MetricType = "gauge"
 	MetricTypeAbsolute MetricType = "absolute"
 )
+
+// IsBuiltinMetric reports whether the given metric key is in the builtin metric namespace
+func IsBuiltinMetric(key string) bool {
+	return strings.HasPrefix(key, builtinMetricPrefix)
+}
 
 // validateValue checks if the supplied metric value fits the requirements
 // of its expected type.
@@ -70,7 +75,7 @@ func ReadMetrics(r io.Reader) (*Metrics, error) {
 		return &metrics, nil
 	}
 	for name, metric := range metrics.Metrics {
-		if strings.HasPrefix(name, builtinMetricsPrefix) {
+		if IsBuiltinMetric(name) {
 			if metric.Type != MetricType("") || metric.Description != "" {
 				return nil, fmt.Errorf("metric %q is using a prefix reserved for built-in metrics: it should not have type or description specification", name)
 			}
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -17,13 +17,18 @@ import (
...
 const (
-	builtinMetricsPrefix = "juju"
+	BuiltinMetricPrefix = "juju"
 
 	// Supported metric types.
 	MetricTypeGauge    MetricType = "gauge"
 	MetricTypeAbsolute MetricType = "absolute"
 )
 
+// IsBuiltinMetric returns true if the metric key provided is in the builtin metric namespace.
```
frankban: Should we add a trivial test for this at this point?
[shrug]

metrics.go
```diff
@@ -17,13 +17,18 @@ import (
 type MetricType string
 
 const (
-	builtinMetricsPrefix = "juju"
+	BuiltinMetricPrefix = "juju"
```
frankban: Do we need to export this as well? Sorry I miss the background for this change.

frankban:
:+1: with a couple of minor questions. Thank you!

metrics.go
rogpeppe: Standard phrasing would be something like:

    // IsBuiltinMetric reports whether the given metric key is in the
    // built-in metric name space.

("built-in" vs "builtin" and "name space" vs "namespace" are still
arguable, but "reports whether" is definitely better IMHO).

metrics.go
rogpeppe: Does this actually need to be exported now?

metrics.go
rogpeppe: But tbh I find it hard to get excited about using this rather than just exporting BuiltinMetricPrefix.

metrics.go
alesstimec: Yes, it does, i think.. can used to compose the built-in metrics name in the core.

rogpeppe:
LGTM but I think that only one thing should be exported - if you need BuiltinMetricPrefix, then axe the strings.HasPrefix wrapper.

metrics.go
alesstimec: +1 to frankban: a simple test would be in place

alesstimec:
LGTM

metrics.go
```diff
@@ -70,7 +75,7 @@ func ReadMetrics(r io.Reader) (*Metrics, error) {
 		return &metrics, nil
 	}
 	for name, metric := range metrics.Metrics {
-		if strings.HasPrefix(name, builtinMetricsPrefix) {
+		if IsBuiltinMetric(name) {
```
tasdomas: This definitely needs a test.

alesstimec:
LGTM

metrics.go
```diff
@@ -17,13 +17,18 @@ import (
...
 const (
-	builtinMetricsPrefix = "juju"
+	builtinMetricPrefix = "juju"
 
 	// Supported metric types.
 	MetricTypeGauge    MetricType = "gauge"
 	MetricTypeAbsolute MetricType = "absolute"
 )
 
+// IsBuiltinMetric reports whether the given metric key is in the builtin metric namespace
```
frankban: Please complete the sentence.

PR #144, opened 2015-08-20
Only copy maps in Workload if they aren't nil.

ericsnowcurrently:
Currently Workload.Copy() turns nil maps into empty-but-allocated maps.  This patch fixes that.


wwitzel3:
LGTM

PR #146, opened 2015-08-26
URLs should refer to entities, rather than charms

makyo:
This is the first step in allowing deployment of bundles from juju with the command line.  This refers to charms and bundles as entities.  Future steps will enable different URL styles and future work in other repos will enable dealing with these references.

### DIFF
--- a/bundledata_test.go
+++ b/bundledata_test.go
@@ -23,29 +23,29 @@ var _ = gc.Suite(&bundleDataSuite{})
 
 const mediawikiBundle = `
 series: precise
-services: 
-    mediawiki: 
+services:
+    mediawiki:
         charm: "cs:precise/mediawiki-10"
         num_units: 1
-        options: 
+        options:
             debug: false
             name: Please set name of wiki
             skin: vector
-        annotations: 
+        annotations:
             "gui-x": 609
             "gui-y": -15
-    mysql: 
+    mysql:
         charm: "cs:precise/mysql-28"
         num_units: 2
         to: [0, mediawiki/0]
-        options: 
+        options:
             "binlog-format": MIXED
             "block-size": 5
             "dataset-size": "80%"
             flavor: distro
             "ha-bindiface": eth0
             "ha-mcastport": 5411
-        annotations: 
+        annotations:
             "gui-x": 610
             "gui-y": 255
         constraints: "mem=8g"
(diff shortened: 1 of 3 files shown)


frankban:
I am +1 on using the more generic "entity" for the error messages.
However, I am not sure that we want to allow the "bundle" schema in `charm.URL` and `charm.Reference`. `bundle:bundle/django-42` is not something I'd like to support, but maybe this is only the initial change and the goal is to support the old `bundle:basket/name` charmworld syntax. Even in this case, I think we should avoid allowing mixed cases, and moreover this syntax is deprecated in quickstart, which also prints out a deprecation warning. There is no basket concept in juju-core, and I believe we should not support `bundle:` URLs here. Fortunately dropping this means only changing a single line of this pull request, IIUC.

These are the forms currently supported in quickstart:
- jujucharms paths like `/u/{user}/{name}/{revision}` in which the user and the revision are both optional, and also the leading slash. Essentially we support copy/pasting of jujucharms URL paths. For instance, we support `u/{user}/{name}`, `{name}/{revison}` and just `{name}`. However, in juju-core, the last case (just `{name}`) is an abbreviation for `cs:bundle/name` rather than a jujucharms form, and this is actually good;
- same as above but as an URL, with the `https://jujucharms.com` prefix, e.g. `https://jujucharms.com/mongodb-cluster/4`. Do we want to still support this?

To those two, we add the usual `cs:~{user}/bundle/{name}-{revision}` syntax, already supported by charm and juju-core. This is the canonical form represented by `charm.Reference` and I'd expect the other two forms to be parsed and converted to `charm.Reference`s. So we have the jujucharms path syntax, the jujucharms URL syntax and the canonical syntax, and their abbreviations, and all those represent a bundle in the charm store. 

What's the plan for bundles not in the charm store?
- we will support `local:bundle/{name}-{revision}` out of the box, and quite automatically, as everything is ready now in `charmsrepo.v1`. This is a bundle directory included in a local Juju repository;
- quickstart supports retrieving a bundle YAML from an arbitrary URL (starting with "http" or "https");
- it also supports deploying local bundle data files (with ".yaml" or ".json" extension).

The last two forms cannot be represented by a bundle reference, because the only thing provided is the YAML (the `charm.BundleData`) which does not include the bundle name or revision. This can be ok because, in order to deploy a bundle, we don't need its reference, we only need its `charm.BundleData`. 

So, to wrap this up, here are the sources from where we want to retrieve a bundle data, so that we can deploy a bundle or generate the corresponding bundle changes for the GUI:
- a charm store canonical URL, like `cs:~frankban/bundle/django-42` or `bundle/mediawiki-single` or just `mediawiki-single`: **already done, no work needed**;
- a local repository canonical URL, like `local:bundle/django`: **basic support already present**;
- a jujucharms path, like `u/who/mediawiki-single/42` or `mediawiki-single/42`: **not implemented yet**;
- a jujucharms URL, like `https://jujucharms.com/mongodb-cluster/4`: **not implemented yet**;
- a remote HTTP or HTTPS URL to a YAML or JSON content: **not implemented yet**;
- a local YAML or JSON file: **not implemented yet**.

Sorry for this, I basically used your pull request to sketch a quick spec for this feature. Hopefully this can be useful as a starting point for discussions. Back to your PR, I am :+1: with removal of the change that allows URLs with "bundle:" schema

makyo:
Thanks @frankban for the input!  I've removed the bundle:foo support, and turned this branch into an update from 'charm' to 'entity' in reference parsing.  Additionally, there's a document surrounding future URL work that you've been copied on.

makyo:
As this is trivial, and only affects text, I'm going to :shipit: Thanks @frankban

PR #148, opened 2015-08-28
Resolve fully qualified storefront URLs

makyo:
This allows passing a storefront URL (eg: https://jujucharms.com/u/charmers/mysql/trusty/28 - see url_test.go for permutations) to parseReference.  This is the first step to supporting v2 charm/bundle URLs

### DIFF
--- a/url.go
+++ b/url.go
@@ -15,10 +15,6 @@ import (
 	"gopkg.in/mgo.v2/bson"
 )
 
-const (
-	charmstoreURL string = "jujucharms.com"
-)
-
 // Location represents a charm location, which must declare a path component
 // and a string representaion.
 type Location interface {
@@ -54,7 +50,6 @@ var ErrUnresolvedUrl error = fmt.Errorf("entity url series is not resolved")
 var (
 	validSeries = regexp.MustCompile("^[a-z]+([a-z0-9]+)?$")
 	validName   = regexp.MustCompile("^[a-z][a-z0-9]*(-[a-z0-9]*[a-z][a-z0-9]*)*$")
-	validStore  = regexp.MustCompile(fmt.Sprintf("^https?://(www\\.)?%s/", charmstoreURL))
 )
 
 // IsValidSeries returns whether series is a valid series in charm URLs.
--- a/url.go
+++ b/url.go
@@ -184,6 +184,9 @@ func parseReference(url string) (*Reference, error) {
 				r.Revision = revision
 			} else {
 				r.Series = parts[0]
+				if !IsValidSeries(r.Series) {
+					return nil, fmt.Errorf("entity URL has invalid series: %q", url)
+				}
 				parts = parts[1:]
 				if len(parts) == 1 {
 					r.Revision, err = strconv.Atoi(parts[0])
--- a/url.go
+++ b/url.go
@@ -160,16 +160,42 @@ func parseReference(url string) (*Reference, error) {
 	// Check if we're dealing with a v1 or v2 URL.
 	// TODO: this is a naive check that only works on fully-qualified URLs;
 	// further work will be required for future URLs
-	urlVersion := 1
-	if strings.Index(url, "http") == 0 {
-		urlVersion = 2
-		if !validStore.MatchString(url) {
-			return nil, fmt.Errorf("entity URL has invalid charmstore: %q", url)
-		}
+	// XXX Get rid of urlVersion, test based on validStore.MatchString
+	if validStore.MatchString(url) {
 		r.Schema = "cs"
 		url = validStore.ReplaceAllString(url, "")
-	}
-	if urlVersion == 1 {
+		parts := strings.Split(url, "/")
+		// Strip any trailing blank parts.
+		if parts[len(parts)-1] == "" {
+			parts = parts[:len(parts)-1]
+		}
+		if parts[0] == "u" {
+			r.User = parts[1]
+			parts = parts[2:]
+		}
+		r.Name = parts[0]
+		parts = parts[1:]
+		r.Revision = -1
+		if len(parts) > 0 {
+			revision, err := strconv.Atoi(parts[0])
+			if err == nil {
+				r.Revision = revision
+			} else {
+				r.Series = parts[0]
+				parts = parts[1:]
+				if len(parts) == 1 {
+					r.Revision, err = strconv.Atoi(parts[0])
+					if err != nil {
+						return nil, fmt.Errorf("entity URL has malformed revision: %q in %q", parts[0], url)
+					}
+				} else {
+					if len(parts) != 0 {
+						return nil, fmt.Errorf("entity URL has invalid form: %q", url)
+					}
+				}
+			}
+		}
+	} else {
 		i := strings.Index(url, ":")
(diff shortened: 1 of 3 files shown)


url.go
```diff
@@ -152,64 +157,106 @@ func ParseReference(url string) (*Reference, error) {
...
-		if r.Schema != "cs" && r.Schema != "local" {
-			return nil, fmt.Errorf("entity URL has invalid schema: %q", url)
+	// Check if we're dealing with a v1 or v2 URL.
+	// TODO: this is a naive check that only works on fully-qualified URLs;
+	// further work will be required for future URLs
+	// XXX Get rid of urlVersion, test based on validStore.MatchString
+	if validStore.MatchString(url) {
+		r.Schema = "cs"
+		url = validStore.ReplaceAllString(url, "")
+		parts := strings.Split(url, "/")
```
frankban: What about something like `parts := strings.Split(strings.Trim(url), "/")` so that we don't need to strip leading/trailing slashes later?

frankban:
:+1: with the change in tests as discussed.
Thanks!

url.go
```diff
@@ -152,64 +157,111 @@ func ParseReference(url string) (*Reference, error) {
...
-	i := strings.Index(url, ":")
-	if i >= 0 {
-		r.Schema = url[:i]
-		if r.Schema != "cs" && r.Schema != "local" {
-			return nil, fmt.Errorf("entity URL has invalid schema: %q", url)
+	// Check if we're dealing with a v1 or v2 URL.
+	// TODO: this is a naive check that only works on fully-qualified URLs;
+	// further work will be required for future URLs
+	// XXX Get rid of urlVersion, test based on validStore.MatchString
+	if validStore.MatchString(url) {
```
rogpeppe: Perhaps split both the arms of this if statement into
two functions (parseNewStyleReference and parseOldStyleReference?).

I'd call parseNewStyleReference iff the url starts with https?: - then
the error messages will be sensible if the url is a new style URL
but not entirely well formed.

url.go
```diff
@@ -152,64 +157,111 @@ func ParseReference(url string) (*Reference, error) {
...
+			if len(parts) < 3 {
+				return nil, fmt.Errorf("entity URL malformed, expecting user and name: %q", url)
+			}
+			r.User = parts[1]
+			parts = parts[2:]
+		}
+		r.Name = parts[0]
+		parts = parts[1:]
+		r.Revision = -1
+		if len(parts) > 0 {
```
rogpeppe: If this was a separate function, you could do:

   if len(parts) == 0 {
       return &r, nil
    }

url.go
```diff
@@ -152,64 +157,111 @@ func ParseReference(url string) (*Reference, error) {
...
+				if !IsValidSeries(r.Series) {
+					return nil, fmt.Errorf("entity URL has invalid series: %q", url)
+				}
+				parts = parts[1:]
+				if len(parts) == 1 {
+					r.Revision, err = strconv.Atoi(parts[0])
+					if err != nil {
+						return nil, fmt.Errorf("entity URL has malformed revision: %q in %q", parts[0], url)
+					}
+				} else {
```
rogpeppe: } else if {

url.go
```diff
@@ -49,6 +53,7 @@ var ErrUnresolvedUrl error = fmt.Errorf("entity url series is not resolved")
 var (
 	validSeries = regexp.MustCompile("^[a-z]+([a-z0-9]+)?$")
 	validName   = regexp.MustCompile("^[a-z][a-z0-9]*(-[a-z0-9]*[a-z][a-z0-9]*)*$")
+	validStore  = regexp.MustCompile(fmt.Sprintf("^https?://(www\\.)?%s/", charmstoreURL))
```
rogpeppe: I'd be tempted use url.Parse to parse
the URL, and then do the rest of the checking on the parts
of the parsed URL.

url.go
```diff
@@ -151,18 +152,35 @@ func ParseReference(url string) (*Reference, error) {
 }
 
 func parseReference(url string) (*Reference, error) {
+	// Check if we're dealing with a v1 or v2 URL.
+	u, err := gourl.Parse(url)
+	if err != nil {
+		return nil, fmt.Errorf("entity URL is not a valid url: %s", url)
```
rogpeppe: perhaps:

    return fmt.Errorf("cannot parse entity URL %q: %v", url, err)

?

url.go
```diff
@@ -151,18 +152,35 @@ func ParseReference(url string) (*Reference, error) {
...
+	}
+	// Shortcut old-style URLs.
+	if u.Opaque != "" {
```
rogpeppe: Cool, I always saw Opaque in passing and didn't know quite what it was. Now I know! Very useful for our purposes.

url.go
```diff
@@ -151,18 +152,35 @@ func ParseReference(url string) (*Reference, error) {
 }
 
 func parseReference(url string) (*Reference, error) {
+	// Check if we're dealing with a v1 or v2 URL.
+	u, err := gourl.Parse(url)
+	if err != nil {
+		return nil, fmt.Errorf("entity URL is not a valid url: %s", url)
+	}
```
rogpeppe: Perhaps:

     if u.FawQuery != "" || u.Fragment != "" || u.User != nil {
          return fmt.Errorf("entity URL %s has unrecognised parts", url)
     }

?

url.go
```diff
@@ -211,6 +226,61 @@ func parseReference(url string) (*Reference, error) {
...
+	return &r, nil
+}
+
+func parseV2Reference(url *gourl.URL) (*Reference, error) {
+	var r Reference
+	r.Schema = "cs"
+	parts := strings.Split(strings.Trim(url.Path, "/"), "/")
+	if len(parts) == 0 {
+		return nil, fmt.Errorf("entity URL malformed, no parts: %q", url)
```
rogpeppe: This can never happen (strings.Split always returns at least one element in the slice)

url.go
```diff
@@ -211,6 +226,61 @@ func parseReference(url string) (*Reference, error) {
...
+	}
+	if parts[0] == "u" {
+		if len(parts) < 3 {
+			return nil, fmt.Errorf("entity URL malformed, expecting user and name: %q", url)
```
rogpeppe: return fmt.Errorf("malformed entity URL %q; expected \"/u/<user>/<name>\"", url)

?

url.go
```diff
@@ -211,6 +226,61 @@ func parseReference(url string) (*Reference, error) {
...
+		}
+		r.User = parts[1]
```
rogpeppe: an idiom that can look nice:

    r.User, parts = parts[1], parts[2:]

(and below)
YMMV.

url.go
```diff
@@ -151,18 +152,35 @@ func ParseReference(url string) (*Reference, error) {
 }
 
 func parseReference(url string) (*Reference, error) {
```
rogpeppe: Please update the doc comment on ParseReference to include the new forms of URL.

rogpeppe:
LGTM thanks!

makyo:
Thanks all

makyo:
$$merge$$

PR #151, opened 2015-09-09
Series list

wallyworld:
The unused Series attribute in charm metadata was a string scalar but has been changed to be a List of supported series.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -185,7 +185,7 @@ type Meta struct {
 	OldRevision int                 `bson:"oldrevision,omitempty"` // Obsolete
 	Categories  []string            `bson:"categories,omitempty"`
 	Tags        []string            `bson:"tags,omitempty"`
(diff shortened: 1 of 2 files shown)


meta.go
```diff
@@ -295,7 +293,7 @@ func (m Meta) GetYAML() (tag string, value interface{}) {
 		Categories  []string                     `yaml:"categories,omitempty"`
 		Tags        []string                     `yaml:"tags,omitempty"`
 		Subordinate bool                         `yaml:"subordinate,omitempty"`
-		Series      string                       `yaml:"series,omitempty"`
+		Series      []string                     `yaml:"series,omitempty"`
```
urosj: Do we allow series to be empty?

wallyworld: Yes, for existing legacy charms.

urosj:
:+1:

PR #155, opened 2015-09-28
Rename series to supported-series

wallyworld:
Rename the Series attribute in Meta to SupportedSeries.
The Series attribute is not used.


rogpeppe:
LGTM thanks

marcoceppi:
Is there a reason for the remane? supported-series is so much longer when series: [] achieves the same effect.

johnsca:
I agree.  "series" seems perfectly reasonable to me and I don't understand why we would change this.

chuckbutler:
series: [] seems succinct enough. I'm not following the logic for the rename either.

PR #158, opened 2015-09-29
Rename SupportedSeries and add json serialisation

wallyworld:
The recent rename of Series to SupportedSeries was rejected, so we rename back.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -173,6 +173,12 @@ func (r Relation) IsImplicit() bool {
 
 // Meta represents all the known content that may be defined
 // within a charm's metadata.yaml file.
+// Note: Series is serialised for backward compatibility
+// as "supported-series" because a previous
+// charm version had an incompatible Series field that
+// was unused in practice but still serialized. This
+// only applies to JSON because Meta has a custom
+// YAML marshaller.
 type Meta struct {
 	Name        string              `bson:"name" json:"name"`
 	Summary     string              `bson:"summary" json:"summary"`
--- a/meta.go
+++ b/meta.go
@@ -174,19 +174,19 @@ func (r Relation) IsImplicit() bool {
 // Meta represents all the known content that may be defined
 // within a charm's metadata.yaml file.
 type Meta struct {
-	Name            string              `bson:"name"`
-	Summary         string              `bson:"summary"`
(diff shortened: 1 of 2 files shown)


meta.go
```diff
@@ -174,19 +174,19 @@ func (r Relation) IsImplicit() bool {
...
+	Description string              `bson:"description" json:"description"`
+	Subordinate bool                `bson:"subordinate" json:"subordinate"`
+	Provides    map[string]Relation `bson:"provides,omitempty" json:"provides,omitempty"`
+	Requires    map[string]Relation `bson:"requires,omitempty" json:"requires,omitempty"`
+	Peers       map[string]Relation `bson:"peers,omitempty" json:"peers,omitempty"`
+	Format      int                 `bson:"format,omitempty" json:"format,omitempty"`
+	OldRevision int                 `bson:"oldrevision,omitempty"` // Obsolete
+	Categories  []string            `bson:"categories,omitempty" json:"categories,omitempty"`
+	Tags        []string            `bson:"tags,omitempty" json:"tag,omitempty"`
+	Series      []string            `bson:"series,omitempty" json:"supported-series,omitempty"`
```
rogpeppe: Perhaps add a comment here:

// Series is serialised for backward compatibility
// as "supported-series" because a previous
// charm version had an incompatible Series field that
// was unused in practice but still serialized. This
// only applies to JSON because Meta has a custom
// YAML marshaler.

?

meta_test.go
```diff
@@ -268,7 +268,7 @@ func (s *MetaSuite) TestSeries(c *gc.C) {
 func (s *MetaSuite) TestInvalidSeries(c *gc.C) {
```
rogpeppe: We should have at least one test that tests we can unmarshal a charm with multiple series.

rogpeppe:
LGTM with a couple of suggestions, thanks.

meta.go
wallyworld: Done

meta_test.go
wallyworld: We do - TestSeries()

wallyworld:
:shipit:

wallyworld:
$$merge$$

PR #159, opened 2015-09-30
Add a supported series helper function

wallyworld:
Add a new SeriesToUse() helper for use by core and charmrepo.

### DIFF
--- a/charm.go
+++ b/charm.go
@@ -41,28 +41,26 @@ func ReadCharm(path string) (charm Charm, err error) {
 	return charm, nil
 }
 
-// SeriesToUse takes a specified series and a list of series supported by a
+// SeriesForCharm takes a requested series and a list of series supported by a
 // charm and returns the series which is relevant.
-// If the specified series is empty, the any default series defined by the
-// charm is used. Any specified series is validated against those supported
-// by the charm.
-func SeriesToUse(series string, supportedSeries []string) (string, error) {
+// If the requested series is empty, then the first supported series is used,
+// otherwise the requested series is validated against the supported series.
+func SeriesForCharm(requestedSeries string, supportedSeries []string) (string, error) {
 	// Old charm with no supported series.
 	if len(supportedSeries) == 0 {
-		return series, nil
+		return requestedSeries, nil
 	}
 	// Use the charm default.
-	if series == "" {
+	if requestedSeries == "" {
 		return supportedSeries[0], nil
 	}
-	// Ensure series is supported.
 	for _, s := range supportedSeries {
-		if s == series {
-			return series, nil
+		if s == requestedSeries {
+			return requestedSeries, nil
 		}
 	}
 	return "", fmt.Errorf(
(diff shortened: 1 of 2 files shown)


charm.go
```diff
@@ -38,3 +40,29 @@ func ReadCharm(path string) (charm Charm, err error) {
 	}
 	return charm, nil
 }
+
+// SeriesToUse takes a specified series and a list of series supported by a
+// charm and returns the series which is relevant.
+// If the specified series is empty, the any default series defined by the
+// charm is used. Any specified series is validated against those supported
+// by the charm.
+func SeriesToUse(series string, supportedSeries []string) (string, error) {
```
axw: Rename the first parameter to "requestedSeries"? I think that would make its purpose a bit more obvious.

charm.go
```diff
@@ -38,3 +40,29 @@ func ReadCharm(path string) (charm Charm, err error) {
 	}
 	return charm, nil
 }
+
+// SeriesToUse takes a specified series and a list of series supported by a
+// charm and returns the series which is relevant.
+// If the specified series is empty, the any default series defined by the
```
axw: s/the any/then any/

Is it really *any*? you're selecting the first one, and in the code you say "charm default". Is the charm default the first value in the list? If so, you should not say "any" here.

charm.go
```diff
@@ -38,3 +40,29 @@ func ReadCharm(path string) (charm Charm, err error) {
...
+// charm is used. Any specified series is validated against those supported
```
axw: s/Any/The/

charm.go
```diff
@@ -38,3 +40,29 @@ func ReadCharm(path string) (charm Charm, err error) {
...
+func SeriesToUse(series string, supportedSeries []string) (string, error) {
+	// Old charm with no supported series.
+	if len(supportedSeries) == 0 {
+		return series, nil
+	}
+	// Use the charm default.
+	if series == "" {
+		return supportedSeries[0], nil
+	}
+	// Ensure series is supported.
```
axw: code is self-evident, comment is unnecessary

axw:
A few doc issues, otherwise LGTM. I don't like the name, but it'll do.

charm.go
wallyworld: "any" needs to be read as "any default"
If the charm does not specify series, the default series is ""
The default is the first series specified, if any are specified.

charm.go
wallyworld: The specified series is optional, so any is appropriate here.

wallyworld: Reworded:

// If the requested series is empty, then any default series defined by the
// charm is used, otherwise the requested series is validated against those
// supported by the charm.

charm.go
axw: If you're saying it's written as intended, then I do not understand it. I would write it as:

    If the requested series is empty, the charm's default series (first item in the supported series list) is used.

Is that correct?

wallyworld:
$$merge$$

PR #160, opened 2015-10-01
Add an error when series cannot be determined

wallyworld:
If a legacy charm does not declare any series, then we error when SeriesForCharm is called without a series.


axw:
Oops. LGTM.

wallyworld:
$$merge$$

PR #161, opened 2015-10-05
Fix for builtin metric validation.

alesstimec:


mattyw:
LGTM - can't believe this wasn't in there before

PR #162, opened 2015-10-08
Add payload classes to charm metadata.

ericsnowcurrently:
This is for the new "Payload Management" feature.


ericsnowcurrently:
See PR #163.

PR #170, opened 2015-11-10
V6 unstable updates

alesstimec:


cmars:
I think this looks ok... lacking context I want to diff against current v6-unstable, just a few min.

cmars:
[LGTM](https://github.com/juju/charm/compare/v6-unstable...alesstimec:v6-unstable-updates)

PR #171, opened 2015-11-11
Parse storage constraints from bundle data

axw:
Add support for storage in bundles, so we can deploy bundles of charms that use the storage feature.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -137,6 +137,10 @@ type ServiceSpec struct {
 	// when creating new machines for units of the service.
 	// This is ignored for units with explicit placement directives.
 	Constraints string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+
+	// Storage holds the constraints for storage to assign
+	// to units of the service.
+	Storage map[string]string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


axw:
@frankban would you mind taking a look please? I'm not sure what the impact of changing the signature of Verify/VerifyWithCharms is, but I think it needs to be done all the same.

bundledata.go
```diff
@@ -303,6 +318,11 @@ func (verifier *bundleDataVerifier) verifyServices() {
 		if err := verifier.verifyConstraints(svc.Constraints); err != nil {
 			verifier.addErrorf("invalid constraints %q in service %q: %v", svc.Constraints, name, err)
 		}
+		for storageName, storageConstraints := range svc.Storage {
```
frankban: Is there any way on which storageName itself can be invalid?

frankban:
Thanks for this branch!
Callers (basically juju-core and the charm store) will need to reflect the change on Verify(WithCharms), which is backward incompatible, but this is an unstable version anyway.
So :+1: with a question about validating storage keys.

bundledata.go
axw: Yes, there is, thanks for picking that up. I've added validation of the storage name too, along with a test.

axw:
$$merge$$

PR #174, opened 2015-11-14
imlpement parsing of Min Juju Version

natefinch:
This adds a very simple minjujuversion key into the metadata.yaml that accepts the standard 1.25.0 style version number for comparison to the juju server version (though the comparison is done in core).  This PR depends on https://github.com/juju/version/pull/1 which moves the juju version parsing code out of juju/juju so that the charm code can use it.


natefinch:
closed in favor of https://github.com/juju/charm/pull/176

PR #176, opened 2015-11-21
implement parsing of Min Juju Version

natefinch:
This adds a very simple min-juju-version key into the metadata.yaml that accepts the standard 1.25.0 style version number for comparison to the juju server version (though the comparison is done in core). This PR depends on juju/version#1 which moves the juju version parsing code out of juju/juju so that the charm code can use it.

### DIFF
--- a/charm.go
+++ b/charm.go
@@ -125,7 +125,3 @@ func CheckMinVersion(ch Charm, jujuVersion version.Number) error {
 type minJujuVersionErr struct {
 	*errors.Err
 }
-
-func minVersionError(minver, jujuver version.Number) error {
-
-}
--- a/charm.go
+++ b/charm.go
@@ -102,9 +102,9 @@ func IsUnsupportedSeriesError(err error) bool {
 	return ok
 }
 
-// isMinVersionErr reports whether the given error indicates a minmum version
+// isMinVersionError reports whether the given error indicates a minmum version
 // too high for this version of juju.
-func IsMinVersionErr(err error) bool {
+func IsMinVersionError(err error) bool {
 	_, ok := err.(minJujuVersionErr)
 	return ok
 }
--- a/charm.go
+++ b/charm.go
@@ -8,7 +8,9 @@ import (
 	"os"
 	"strings"
 
+	"github.com/juju/errors"
 	"github.com/juju/loggo"
+	"github.com/juju/version"
 )
 
 var logger = loggo.GetLogger("juju.charm")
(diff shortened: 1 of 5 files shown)


charm_test.go
```diff
@@ -102,6 +103,30 @@ func (s *CharmSuite) IsMissingSeriesError(c *gc.C) {
...
+	meta *charm.Meta
+}
+
+func (f fakeCharm) Meta() *charm.Meta {
+	return f.meta
+}
+
+func (s *CharmSuite) TestIsMinVersionError(c *gc.C) {
+	err := charm.MinVersionError()
+	c.Assert(charm.IsMinVersionError(err), jc.IsTrue)
```
fwereade: nice to make these Checks so they don't stop the test

meta.go
```diff
@@ -294,29 +307,31 @@ func (m Meta) GetYAML() (tag string, value interface{}) {
...
+		Description    string                       `yaml:"description"`
+		Provides       map[string]marshaledRelation `yaml:"provides,omitempty"`
+		Requires       map[string]marshaledRelation `yaml:"requires,omitempty"`
+		Peers          map[string]marshaledRelation `yaml:"peers,omitempty"`
+		Categories     []string                     `yaml:"categories,omitempty"`
+		Tags           []string                     `yaml:"tags,omitempty"`
+		Subordinate    bool                         `yaml:"subordinate,omitempty"`
+		Series         []string                     `yaml:"series,omitempty"`
+		Terms          []string                     `yaml:"terms,omitempty"`
+		MinJujuVersion *version.Number              `yaml:"min-juju-version,omitempty"`
```
fwereade: not really clear why this is a pointer -- surely the zero version.Number is perfectly suitable, and less likely to induce potential panics when not initialized...

meta.go
```diff
@@ -277,9 +279,20 @@ func ReadMeta(r io.Reader) (meta *Meta, err error) {
 	meta.Series = parseStringList(m["series"])
 	meta.Storage = parseStorage(m["storage"])
 	meta.PayloadClasses = parsePayloadClasses(m["payloads"])
+
+	if v, ok := m["min-juju-version"].(string); ok {
```
fwereade: surely missing is fine, and valid is fine, but I think invalid data should be considered an error

charm.go
```diff
@@ -93,3 +95,27 @@ func IsUnsupportedSeriesError(err error) bool {
...
+// too high for this version of juju.
+func IsMinVersionError(err error) bool {
+	_, ok := err.(minJujuVersionErr)
+	return ok
+}
+
+// CheckMinVersion reports an error that reports true from IsMinVersionErr if
+// the given charm has a minimum juju version specified that is higher than the
+// given juju version number.
+func CheckMinVersion(ch Charm, jujuVersion version.Number) error {
```
fwereade: this doesn't quite feel like the right place for the error -- and the reference to an "environment" below makes me doubly sure (charm shouldn't need to reference the concept at all). Please move this stuff inside juju/juju

meta.go
natefinch: Oh yeah, totally.  Wasn't thinking.

meta.go
natefinch: good point.

natefinch:
moving to a feature branch

PR #184, opened 2015-12-15
Adjust resources code.

ericsnowcurrently:
This patch makes a few adjustments needed for the resources feature:

* Make resource.Type an int enum.
* Add resource.Origin and Resource.Origin.
* Return an error from resource.ParseType().  Returning a bool (and then ignoring it) wasn't very helpful.

This patch also percolates the parsing error up the stack.  The pattern of ignoring errors during metadata parsing relies on parse results causing appropriate errors during validation.  This approach is fragile and easily resolved by supporting error returns from parsing.

### DIFF
--- a/resource/origin.go
+++ b/resource/origin.go
@@ -1,0 +1,50 @@
+// Copyright 2015 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package resource
+
+import (
+	"github.com/juju/errors"
+)
+
+// These are the valid resource origins.
+const (
+	originUnknown Origin = iota
+	OriginUpload
+	OriginStore
+)
+
+var origins = map[Origin]string{
+	OriginUpload: "upload",
+	OriginStore:  "store",
+}
+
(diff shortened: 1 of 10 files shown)


resource/origin.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// These are the valid resource origins.
+const (
+	originUnknown Origin = iota
+	OriginUpload
+	OriginStore
+)
+
+var origins = map[Origin]string{
```
kat-co: Complete developer preference here, but I believe the stringer @natefinch demonstrated in the last sprint is *a* standard way to turn iotas into strings. Use this? If not, that's totally fine.

kat-co:
Overall LGTM

resource/origin.go
ericsnowcurrently: Doing that makes parsing more of a pain.

wwitzel3:
LGTM

ericsnowcurrently:
$$merge$$

PR #186, opened 2016-01-06
Add endpoint binding parsing to charm bundles

dooferlad:
This change adds the ability to bind an endpoint to a space inside a charm bundle, for example:
```yaml
services:
    wordpress:
        charm: wordpress
        num_units: 1
        bindings:
            db: db
            url: public
```
This will request that wordpress has its database endpoint (db) attached to the db space and the url endpoint attached to the public space.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -463,34 +463,22 @@ func (verifier *bundleDataVerifier) verifyEndpointBindings() {
 	for name, svc := range verifier.bd.Services {
 		charm, ok := verifier.charms[name]
 		// Only thest the ok path here because the !ok path is tested in verifyServices
-		if ok {
-			for BoundInterface := range svc.EndpointBindings {
-				var ok bool
+		if !ok {
+			continue
+		}
+		for endpoint, space := range svc.EndpointBindings {
+			_, matchedProvides := charm.Meta().Provides[endpoint]
+			_, matchedRequires := charm.Meta().Requires[endpoint]
+			_, matchedPeers := charm.Meta().Peers[endpoint]
 
-				for _, relation := range charm.Meta().Provides {
-					if relation.Interface == BoundInterface {
-						ok = true
-						break
-					}
-				}
-
-				if !ok {
-					for _, relation := range charm.Meta().Requires {
-						if relation.Interface == BoundInterface {
-							ok = true
-							break
-						}
-					}
-				}
-
-				if !ok {
-					verifier.addErrorf(
-						"service %s wants to bind to interface %s, "+
-							"which isn't provided or required by charm",
-						name, BoundInterface)
-				}
+			if !(matchedProvides || matchedRequires || matchedPeers) {
+				verifier.addErrorf(
+					"service %s wants to bind to endpoint %s to space %s, "+
+						"which isn't provided or required by charm",
+					name, endpoint, space)
 			}
 		}
+
 	}
 }
 
--- a/bundledata.go
+++ b/bundledata.go
@@ -141,6 +141,9 @@ type ServiceSpec struct {
 	// Storage holds the constraints for storage to assign
 	// to units of the service.
 	Storage map[string]string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+
+	// EndpointBindings maps how endpoints are bound to spaces
+	EndpointBindings map[string]string `bson:"bindings,omitempty" json:"bindings,omitempty" yaml:"bindings,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


bundledata.go
```diff
@@ -455,6 +459,41 @@ func (verifier *bundleDataVerifier) verifyRelations() {
 	}
 }
 
+func (verifier *bundleDataVerifier) verifyEndpointBindings() {
+	for name, svc := range verifier.bd.Services {
+		charm, ok := verifier.charms[name]
+		// Only thest the ok path here because the !ok path is tested in verifyServices
+		if ok {
```
dimitern: To reduce some nesting, how about:
```go
if !ok {
  continue
}
```

bundledata.go
```diff
@@ -141,6 +141,9 @@ type ServiceSpec struct {
 	// Storage holds the constraints for storage to assign
 	// to units of the service.
 	Storage map[string]string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+
+	// EndpointBindings maps how endpoints are bound to spaces
+	EndpointBindings map[string]string `bson:"bindings,omitempty" json:"bindings,omitempty" yaml:"bindings,omitempty"`
```
dimitern: As discussed on IRC I'm surprised Storage ... `yaml ",omitempty"` works correctly for "storage:", but if it does, let's drop the "bindings" in all 3 cases for consistency.

bundledata.go
```diff
@@ -455,6 +459,41 @@ func (verifier *bundleDataVerifier) verifyRelations() {
 	}
 }
 
+func (verifier *bundleDataVerifier) verifyEndpointBindings() {
+	for name, svc := range verifier.bd.Services {
+		charm, ok := verifier.charms[name]
+		// Only thest the ok path here because the !ok path is tested in verifyServices
+		if ok {
+			for BoundInterface := range svc.EndpointBindings {
```
dimitern: I think this needs to be called "endpoint", as it's the relation name we're binding to a space, not the interface of the relation. E.g. with a metadata like this:
```yaml
provides:
    db:
      interface: mysql
```
and bundle.yaml having:
```yaml
...
    bindings:
        db: storage
...
```
"db" (not "mysql" - the interface) is bound to space "storage".

bundledata.go
```diff
@@ -455,6 +459,41 @@ func (verifier *bundleDataVerifier) verifyRelations() {
...
+				var ok bool
+
+				for _, relation := range charm.Meta().Provides {
+					if relation.Interface == BoundInterface {
+						ok = true
+						break
+					}
+				}
+
+				if !ok {
```
dimitern: There needs to be a similar check for charm.Meta().Peers as well.

bundledata.go
```diff
@@ -455,6 +459,41 @@ func (verifier *bundleDataVerifier) verifyRelations() {
...
+							ok = true
+							break
+						}
+					}
+				}
+
+				if !ok {
+					verifier.addErrorf(
+						"service %s wants to bind to interface %s, "+
+							"which isn't provided or required by charm",
```
dimitern: I think it will be helpful to include the space name in the error message.
s/to interface %s/endpoint %q to space %q/
s/isn't provided or required by charm/isn't among the defined charm relations/ ?

dimitern:
It's nearly there, just a few issues need fixing.

bundledata.go
dooferlad: We want the YAML field to be call "bindings" but, internally, we always have used EndpointBindings, hence the explicit name.

dimitern: D'oh. Sure, that's what I asked for anyway :) Please, ignore my previous comment about it.

bundledata.go
```diff
@@ -455,6 +459,29 @@ func (verifier *bundleDataVerifier) verifyRelations() {
...
+		}
+		for endpoint, space := range svc.EndpointBindings {
+			_, matchedProvides := charm.Meta().Provides[endpoint]
+			_, matchedRequires := charm.Meta().Requires[endpoint]
+			_, matchedPeers := charm.Meta().Peers[endpoint]
+
+			if !(matchedProvides || matchedRequires || matchedPeers) {
+				verifier.addErrorf(
+					"service %s wants to bind to endpoint %s to space %s, "+
+						"which isn't provided or required by charm",
```
dimitern: Strictly speaking, "provided or required" is misleading, as it also checks "peers". So how about "isn't defined by the charm" ?

dooferlad:
$$merge$$

PR #188, opened 2016-01-08
Fix GenerateFingerprint() and add ParseFingerprint().

ericsnowcurrently:
This patch changes GenerateFingerprint to take a reader rather than bytes.  This aligns better with how it will be used.

ParseFingerprint() is added to provide a roundtrip for Fingerprint.String().  This is useful when Fingerprint.String() is used to serialize.  Then ParseFingerprint() can be used to deserialize.

### DIFF
--- a/resource/fingerprint_test.go
+++ b/resource/fingerprint_test.go
@@ -57,6 +57,22 @@ func (s *FingerprintSuite) TestNewFingerprintTooBig(c *gc.C) {
 	c.Check(err, gc.ErrorMatches, `.*too big.*`)
 }
 
+func (s *FingerprintSuite) TestParseFingerprintOkay(c *gc.C) {
+	_, expected := newFingerprint(c, "spamspamspam")
+
+	fp, err := resource.ParseFingerprint(expected)
+	c.Assert(err, jc.ErrorIsNil)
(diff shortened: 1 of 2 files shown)


resource/fingerprint_test.go
```diff
@@ -56,10 +57,27 @@ func (s *FingerprintSuite) TestNewFingerprintTooBig(c *gc.C) {
...
+
+func (s *FingerprintSuite) TestParseFingerprintNonHex(c *gc.C) {
+	_, err := resource.ParseFingerprint("XYZ") // not hex
+
+	c.Check(err, gc.ErrorMatches, `.*odd length hex string.*`)
+}
+
 func (s *FingerprintSuite) TestGenerateFingerprint(c *gc.C) {
 	expected, _ := newFingerprint(c, "spamspamspam")
+	data := bytes.NewBufferString("spamspamspam")
```
natefinch: FYI, strings.NewReader is more appropriate here, it's the string->io.Reader that you're looking for.

natefinch:
LGTM, though it'll break a bunch of stuff in our code... perhaps we don't actually use this until after the demo?

ericsnowcurrently:
$$merge$$

PR #191, opened 2016-03-01
Added support for ExtraBindings in Meta

dimitern:
Added an optional `extra-bindings` metadata section, with the following format:
```yaml
...
extra-bindings:
   name1:
   other-name2:
   ...
```
None of the names in the extra-bindings map can match existing provided/required/peer relation
names, as relations are already bindable. Extra bindings provide a way to declare named endpoints,
which the charm needs (bound) addresses for, but which are not relations otherwise.

Extra bindings will be accepted in $ juju deploy --bind ... when declared, in a follow-up.

Also added a Meta.CombinedRelations() helper, fixing http://pad.lv/1520623 bug as a result.

### DIFF
--- a/extra_bindings.go
+++ b/extra_bindings.go
@@ -1,0 +1,117 @@
+// Copyright 2016 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package charm
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/juju/schema"
+)
+
+// ExtraBinding represents a bindable endpoint which is not also a relation.
+type ExtraBinding struct {
+	Name string `bson:"name" json:"Name"`
+}
+
+// When specified, the "extra-bindings" section in the metadata.yaml
+// should have the following format:
+//
+// extra-bindings:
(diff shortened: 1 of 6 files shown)


extra_bindings.go
```diff
@@ -0,0 +1,69 @@
...
+
+func parseMetaExtraBindings(data interface{}) (map[string]ExtraBinding, error) {
+	if data == nil {
+		return nil, nil
+	}
+
+	bindingsMap := data.(map[interface{}]interface{})
+	result := make(map[string]ExtraBinding)
+	for name, _ := range bindingsMap {
+		stringName := name.(string)
```
jameinel: is it ok to do blind casts like this? Doesn't that cause a panic if it isn't a string?

dimitern: Yes, because parseMetaExtraBindings (like other similar helpers for resources, payloads, storage, etc.) are called once the top-level charm schema has been Coerce()d.

jameinel:
I had one comment, but otherwise LGTM

dimitern:
Thanks for the reviews! All issues fixed, setting to $$merge$$.

PR #194, opened 2016-03-08
Merging features necessary for Juju Min Version

kat-co:
This introduces a dependency on github.com/juju/version.


meta_test.go
```diff
@@ -357,6 +358,43 @@ func (s *MetaSuite) TestInvalidSeries(c *gc.C) {
...
+
+// TestInvalidSeries ensures that invalid series values cause a parse error
+// when specified in the charm metadata.
+func (s *MetaSuite) TestInvalidMinJujuVersion(c *gc.C) {
+	_, err := charm.ReadMeta(strings.NewReader(dummyMetadata + "\nmin-juju-version: invalid-version"))
+
+	c.Check(err, gc.ErrorMatches, `invalid version "invalid-version"`)
+}
+
+// TestInvalidSeries ensures that invalid series values cause a parse error
```
wallyworld: comment need fixing

meta_test.go
```diff
@@ -357,6 +358,43 @@ func (s *MetaSuite) TestInvalidSeries(c *gc.C) {
...
+		c.Assert(meta.MinJujuVersion, gc.Equals, ver)
+	}
+}
+
+// TestInvalidSeries ensures that invalid series values cause a parse error
+// when specified in the charm metadata.
+func (s *MetaSuite) TestInvalidMinJujuVersion(c *gc.C) {
+	_, err := charm.ReadMeta(strings.NewReader(dummyMetadata + "\nmin-juju-version: invalid-version"))
+
+	c.Check(err, gc.ErrorMatches, `invalid version "invalid-version"`)
```
wallyworld: can the error message be 'invalid min-version "foo"' so the user knows what attribute is wrong

wallyworld:
LGTM with a couple of trivial fixes

wallyworld:
If we stick with having the attr as a version.Number we'd add schema support for it. May a todo would be good

natefinch:
LGTM

kat-co:
$$merge$$

PR #196, opened 2016-03-09
Allow bundle charms as paths

wallyworld:
We allow bundle.yaml files to contain charm entries where the value is a relative or absolute path. This is used when deploying local bundles where the charms are located in a directory along with the bundle file.


axw:
LGTM

wallyworld:
$$merge$$

PR #197, opened 2016-03-10
Add series attribute to bundles

wallyworld:
Bundle allow a series to be specified. It is an error to specify a series and a charm store charm. Series only applies to local charms.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -68,6 +68,11 @@ type ServiceSpec struct {
 	// use for the given service.
 	Charm string
 
+	// Series is the series to use when deploying the charm
+	// if the charm does not specify a default or the default
+	// is not the desired value.
+	Series string `yaml:",omitempty" json:",omitempty"`
+
 	// NumUnits holds the number of units of the
 	// service that will be deployed.
 	//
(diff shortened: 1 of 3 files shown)


bundledata.go
```diff
@@ -68,6 +68,11 @@ type ServiceSpec struct {
 	// use for the given service.
 	Charm string
 
+	// Series is the series to use when deploying the charm
```
axw: worth mentioning here that the series may not be specified for cs: charms?

bundledata.go
```diff
@@ -364,9 +371,18 @@ func (verifier *bundleDataVerifier) verifyServices() {
...
 				}
 			}
-		} else if _, err := ParseURL(svc.Charm); err != nil {
+		} else if curl, err = ParseURL(svc.Charm); err != nil {
 			verifier.addErrorf("invalid charm URL in service %q: %v", name, err)
 		}
+
+		// Check the series.
+		if curl != nil && svc.Series != "" {
+			verifier.addErrorf("service %q declares both a series and a charm store URL", name)
```
axw: "store" in the message may not be accurate in the future, if we add additional URL schemes

axw:
LGTM with a couple of quibbles

bundledata.go
wallyworld: done

bundledata.go
wallyworld: done

wallyworld:
$$merge$$

PR #198, opened 2016-03-10
Fix bundle charm series validation

wallyworld:
The charm URL in a bundle can include a series so long as it matches any value of the series attribute.


axw:
LGTM

wallyworld:
$$merge$$

PR #199, opened 2016-03-14
Allow negative resource revisions.

ericsnowcurrently:
Negative resource revisions allow us to indicate that no revision is set.  This is what we already do for charm revisions.


natefinch:
LGTM

ericsnowcurrently:
$$merge$$

PR #204, opened 2016-03-30
Reserve "juju" and "juju-*" actions

axw:
Just as we do for relation names, reserve the
action names "juju", and anything starting with
"juju-".

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -109,8 +109,11 @@ func ReadActionsYaml(r io.Reader) (*Actions, error) {
 		if valid := actionNameRule.MatchString(name); !valid {
 			return nil, fmt.Errorf("bad action name %s", name)
 		}
-		if reservedName(name) {
-			return nil, fmt.Errorf("reserved action name %s", name)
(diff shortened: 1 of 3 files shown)


actions.go
```diff
@@ -109,6 +109,9 @@ func ReadActionsYaml(r io.Reader) (*Actions, error) {
 		if valid := actionNameRule.MatchString(name); !valid {
 			return nil, fmt.Errorf("bad action name %s", name)
 		}
+		if reservedName(name) {
+			return nil, fmt.Errorf("reserved action name %s", name)
```
alesstimec: i'd rather see a more useful message.. perhaps: `action name may not start with a reserved prefix "juju"`

alesstimec:
LGTM with a nitpick

actions.go
axw: I've formatted/worded it a little differently, but kept with the spirit of making the error message human-readable.

axw:
$$merge$$

PR #208, opened 2016-05-16
metrics: Reading a metrics file that doesn't define any metrics should be an error

mattyw:
This is ok to review, but @cmars @alesstimec we should discuss it first


tasdomas:
LGTM

PR #209, opened 2016-05-20
Use names.IsValidPayload() to validate the class name.

ericsnowcurrently:
The class name should align with the valid payload ID defined for payload tags.  Otherwise the name cannot be turned into a tag.


kat-co:
LGTM

ericsnowcurrently:
$$merge$$

PR #211, opened 2016-05-27
Handle nil BundleData when unmarshalling from bson

wallyworld:
A small fix for bson unmarshall to properly return SetZero when there's no data.

### DIFF
--- a/bundledata_test.go
+++ b/bundledata_test.go
@@ -306,6 +306,12 @@ func (s *bundleDataSuite) TestBsonMarshall(c *gc.C) {
 	}
 }
 
+func (s *bundleDataSuite) TestBsonNilData(c *gc.C) {
+	var bd charm.BundleData
+	err := bd.SetBSON(bson.Raw{10, []byte(nil)})
+	c.Assert(err, gc.Equals, bson.SetZero)
+}
+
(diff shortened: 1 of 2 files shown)


bundledata_test.go
```diff
@@ -306,6 +306,12 @@ func (s *bundleDataSuite) TestBsonMarshall(c *gc.C) {
 	}
 }
 
+func (s *bundleDataSuite) TestBsonNilData(c *gc.C) {
+	var bd charm.BundleData
+	err := bd.SetBSON(bson.Raw{10, []byte(nil)})
+	c.Assert(err, gc.Equals, bson.SetZero)
+}
```
axw: Can you make the test more about the expected behaviour? The bson.SetZero malarkey is more about the bson internals, rather than what we expect of the de/serialisation.

i.e. marshal an empty BundleData, and unmarshal into a *BundleData and check that it's nil on the way back in.

axw:
LGTM with a higher-level test.

bundledata_test.go
wallyworld: Fair point, added better test

wallyworld:
$$merge$$

PR #215, opened 2016-07-15
meta: Added support for parsing terms with owners

mattyw:


meta_test.go
```diff
@@ -56,14 +56,45 @@ func (s *MetaSuite) TestReadMetaVersion2(c *gc.C) {
...
+		"owner/foo-bar",
+		"owner/foo_bar",
+		"own-er/foobar",
+		"own_er/foobar",
+	}
+
+	invalid := []string{
+		"27",
+		"owner/Foo-bar",
+		"owner/foo/foobar",
```
cmars: Can we also add:
```
/
/1
//
//2
```

cmars: Also, `foo/003`

cmars: And, `ibm/j9-jvm/2`

cmars:
Looking good, some suggestions for additional test cases.

meta_test.go
```diff
@@ -56,14 +56,55 @@ func (s *MetaSuite) TestReadMetaVersion2(c *gc.C) {
...
+		"owner/Foo-bar",
+		"owner/foobar/27",
+		"owner/foobar",
+		"owner/foo-bar",
+		"owner/foo_bar",
+		"own-er/foobar",
+		"own_er/foobar",
+		"ibm/j9-jvm/2",
+		"term_123-23aAf/1",
+		"@les/term/1",
```
cmars: I don't agree with this being a valid owner if it would not be allowed in a charm URL.

meta.go
```diff
@@ -241,16 +241,53 @@ func parseStringList(list interface{}) []string {
 	return result
 }
 
-var termNameRE = regexp.MustCompile("^[a-z]+([a-z0-9-]+)(/[0-9]+)?$")
+var (
+	revSnippet  = `(?:\/(\d+))`
+	termSnippet = `[a-zA-Z][\w-]`
+	//termNameRE  = regexp.MustCompile(fmt.Sprintf(`^(?:(%s+)\/)?((?:%s)+)%s?$`, termSnippet, termSnippet, revSnippet))
+	termNameRE = regexp.MustCompile(`^(?:([^\/\\][\w-]+)\/)?([a-zA-Z][\w-]+)(?:\/(\d+))?$`)
```
cmars: Charm URLs validate the user component with names.IsValidUser. I think term ID validation should do the same for consistency.

A regex is fine to validate the general syntactical structure (slash-delimited fields with some basic constraints, etc), but it might be too hard to read and understand to perform _all_ the validation in the regex.

cmars:
Comments above. It looks like Launchpad usernames aren't always suitable for charm URLs. I wonder if @alesstimec is able to push/publish charms to his own user namespace?

I'd rather that we keep parity with charm URL parsing for this PR and then solve the "allowed characters in usernames" problem consistently in [names.IsValidUser](https://godoc.org/github.com/juju/names#IsValidUser) as this will probably cause trouble for CS and Juju anyway.

mattyw:
@cmars PTAL. I wasn't sure if the default tenant (non specified) should be "cs" or not

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+	if !validTermName.MatchString(t.Name) {
+		return fmt.Errorf("wrong term name format %q", t.Name)
+	}
+	if t.Revision < 0 {
+		return fmt.Errorf("negative term revision")
 	}
 	return nil
 }
 
+// ParseTerm takes a term URI as a string and parses it into a Term.
```
rogpeppe: I wish we wouldn't call this kind of thing a "URI", as it's not really anything of the sort, other than it has a colon and some slashes in.

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+// A complete term is in the form:
+// tenant:owner/name/revision
+// This function accpets partial forms of this.
```
rogpeppe: s/accpets/accepts/

And maybe:

    This function accepts partially-specified identifiers in one of the following forms:

?

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+	if t := strings.Split(s, ":"); len(t) == 2 {
+		tenant = t[0]
+		termid = t[1]
+	}
+
+	tokens := strings.Split(termid, "/")
+	var term Term
+	switch len(tokens) {
+	case 1:
+		term = Term{tenant, "", tokens[0], 0}
```
rogpeppe: I think the code would be more readable as:

    term = Term{
        Tenant: tenant,
        Name: tokens[0],
    }

and in the initializers below too.

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+	termid := s
+	if t := strings.Split(s, ":"); len(t) == 2 {
+		tenant = t[0]
+		termid = t[1]
+	}
+
+	tokens := strings.Split(termid, "/")
+	var term Term
+	switch len(tokens) {
+	case 1:
```
rogpeppe: // name

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+		term = Term{tenant, "", tokens[0], 0}
+	case 2:
```
rogpeppe: // owner/name or name/283

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+// Typically in one of the following forms:
+// name
+// owner/name
+// owner/name/27 # Revision 27
+// name/283 # Revision 283
+// cs:owner/name # Tenant cs
+func ParseTerm(s string) (*Term, error) {
+	tenant := ""
+	termid := s
+	if t := strings.Split(s, ":"); len(t) == 2 {
```
rogpeppe: If there's an extra colon in there, we won't get a great error.
I'd suggest either doing SplitN(n, ":", 2) or returning an error if len(t) > 2.

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+	case 1:
+		term = Term{tenant, "", tokens[0], 0}
+	case 2:
+		termRevision, err := strconv.Atoi(tokens[1])
+		if err != nil {
+			term = Term{tenant, tokens[0], tokens[1], 0}
+		} else {
+			term = Term{tenant, "", tokens[0], termRevision}
+		}
+	case 3:
```
rogpeppe: // owner/name/27

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+		termRevision, err := strconv.Atoi(tokens[2])
+		if err != nil {
+			return nil, errors.Trace(err)
```
rogpeppe: The error from Atoi is not very useful. I'd suggest returning a more informative error:

    return nil, errors.Errorf("invalid revision number in terms id %q", s)

?

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+		}
+		term = Term{tenant, tokens[0], tokens[1], termRevision}
+	default:
+		return nil, errors.Errorf("unknown term format %q", s)
```
rogpeppe: s/term/terms id/ ?

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+type Term struct {
+	Tenant   string
+	Owner    string
+	Name     string
+	Revision int
+}
+
+// Validate returns an error if the Term contains invalid data.
+func (t *Term) Validate() error {
+	if t.Tenant != "" && t.Tenant != "cs" {
```
rogpeppe: Why the special case for "cs" here? ISTM that it's redundant, and why do we care about that value? Do terms ids live in the same namespace as charm ids?

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
+		if !validTermName.MatchString(t.Tenant) {
```
rogpeppe: Why use validTermName to validate the tenant? It's not really the same thing, is it?

meta.go
```diff
@@ -241,16 +242,90 @@ func parseStringList(list interface{}) []string {
...
-	match := termNameRE.FindStringSubmatch(s)
-	if match == nil {
-		return fmt.Errorf("invalid term name %q: must match %s", s, termNameRE.String())
+// Term represents a single term URI. The term can either be owned
+// or "public" (meaning there is no owner).
+type Term struct {
+	Tenant   string
+	Owner    string
+	Name     string
+	Revision int
```
rogpeppe: Perhaps document these fields?
In particular, it might be worth documenting that revisions start from 1 (unlike charm ids) so that it's clear that a zero Revision implies that the revision has not been set in the terms id.

meta_test.go
```diff
@@ -56,48 +56,173 @@ func (s *MetaSuite) TestReadMetaVersion2(c *gc.C) {
...
+		"foo/003",
+		"owner/Foo-bar",
+		"owner/foobar/27",
+		"owner/foobar",
+		"owner/foo-bar",
+		"owner/foo_bar",
+		"own-er/foobar",
+		"ibm/j9-jvm/2",
+		"term_123-23aAf/1",
+		"cs:foobar/27",
```
rogpeppe: Let's have some more tenant forms too.

meta_test.go
```diff
@@ -56,48 +56,173 @@ func (s *MetaSuite) TestReadMetaVersion2(c *gc.C) {
...
+
+	invalid := []string{
+		"/",
+		"/1",
+		"//",
+		"//2",
+		"27",
+		"owner/foo/foobar",
+		"@les/term/1",
+		"own_er/foobar",
```
rogpeppe: I'd think it would be good to have the actual error strings in the invalid tests, so we can see whether the error messages look reasonable.

Also, I'd suggest some more invalid forms (multiple colons, invalid tenants, single hyphens, etc)

rogpeppe: ... but I see that you do have more detailed tests below though. I wonder whether you could just reuse the test data used by TestParseTerms for this test too.

meta_test.go
```diff
@@ -56,48 +56,173 @@ func (s *MetaSuite) TestReadMetaVersion2(c *gc.C) {
...
+func (s *MetaSuite) TestParseTerms(c *gc.C) {
+	tests := []struct {
+		about       string
+		term        string
+		expectError string
+		expectTerm  charm.Term
+	}{{
+		about:      "valid term",
+		term:       "term/1",
+		expectTerm: charm.Term{"", "", "term", 1},
```
rogpeppe: Again, I think these literals read more easily when they're tagged
with field names, so we don't have to remember the field order.

meta.go
rogpeppe: Also, I'm not sure that singular "Term" is the right thing. What we're talking about, AIUI is an identifier that refers to a set of terms and conditions. The singular form of "term" seems to imply that we're talking about a single term, which isn't true. How about standardising on "TermsId" throughout?

meta_test.go
```diff
@@ -56,48 +56,173 @@ func (s *MetaSuite) TestReadMetaVersion2(c *gc.C) {
...
+	for i, test := range tests {
+		c.Logf("running test %v: %v", i, test.about)
+		term, err := charm.ParseTerm(test.term)
+		if test.expectError == "" {
+			c.Check(err, jc.ErrorIsNil)
+			c.Check(term, gc.DeepEquals, &test.expectTerm)
 		} else {
-			c.Assert(err, gc.ErrorMatches, test.expectError)
+			c.Check(err, gc.ErrorMatches, test.expectError)
 		}
```
rogpeppe: c.Check(term, gc.DeepEquals, charm.Term{})

rogpeppe:
Looks reasonable modulo a bunch of comments, thanks. What do you think?

cmars:
LGTM

mattyw:
$$merge$$

meta.go
```diff
@@ -241,16 +242,110 @@ func parseStringList(list interface{}) []string {
...
+		termid = t[1]
+	}
+
+	tokens := strings.Split(termid, "/")
+	var term TermID
+	switch len(tokens) {
+	case 1: // "name"
+		term = TermID{
+			Tenant:   tenant,
+			Owner:    "",
```
rogpeppe: I'd omit the zero-valued fields. That's part of why the tagged-literal form is nicer.

meta.go
```diff
@@ -241,16 +242,110 @@ func parseStringList(list interface{}) []string {
...
 }
 
-var termNameRE = regexp.MustCompile("^[a-z]+([a-z0-9-]+)(/[0-9]+)?$")
+var validTermName = regexp.MustCompile(`^[a-zA-Z][\w-]+$`)
 
-func checkTerm(s string) error {
-	match := termNameRE.FindStringSubmatch(s)
-	if match == nil {
-		return fmt.Errorf("invalid term name %q: must match %s", s, termNameRE.String())
+// TermID represents a single term id. The term can either be owned
```
rogpeppe: As discussed online, I still think that "terms" is better to use throughout. "term" isn't a thing and has other connotations in a computery context.

meta.go
```diff
@@ -241,16 +242,110 @@ func parseStringList(list interface{}) []string {
...
+// or "public" (meaning there is no owner).
+type TermID struct {
+	Tenant   string
+	Owner    string
+	Name     string
+	Revision int
```
rogpeppe: Document that revisions start from 1 and that 0 means unset?

PR #216, opened 2016-07-21
meta: Added Terms Stringer & parsing bug fix

mattyw:
meta: Fixed a bug where terms parsing was happening after being checked.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -244,9 +244,11 @@ func parseStringList(list interface{}) []string {
 
 var validTermName = regexp.MustCompile(`^[a-zA-Z][\w-]+$`)
 
-// TermID represents a single term id. The term can either be owned
+// TermsId represents a single term id. The term can either be owned
 // or "public" (meaning there is no owner).
-type TermID struct {
+// The Revision starts at 1. Therefore a value of 0 means the revision
+// is unset.
+type TermsId struct {
 	Tenant   string
 	Owner    string
 	Name     string
@@ -254,7 +256,7 @@ type TermID struct {
 }
 
 // Validate returns an error if the Term contains invalid data.
-func (t *TermID) Validate() error {
+func (t *TermsId) Validate() error {
 	if t.Tenant != "" && t.Tenant != "cs" {
 		if !validTermName.MatchString(t.Tenant) {
 			return fmt.Errorf("wrong term tenant format %q", t.Tenant)
@@ -272,6 +274,32 @@ func (t *TermID) Validate() error {
 	return nil
 }
 
+// String returns the term in canonical form.
+// This would be one of:
+//   tenant:owner/name/revision
+//   tenant:name
+//   owner/name/revision
+//   owner/name
+//   name/revision
+//   name
+func (t *TermsId) String() string {
+	id := make([]byte, 0, len(t.Tenant)+1+len(t.Owner)+1+len(t.Name)+4)
+	if t.Tenant != "" {
+		id = append(id, t.Tenant...)
+		id = append(id, ':')
+	}
+	if t.Owner != "" {
+		id = append(id, t.Owner...)
+		id = append(id, '/')
+	}
+	id = append(id, t.Name...)
+	if t.Revision != 0 {
+		id = append(id, '/')
+		id = strconv.AppendInt(id, int64(t.Revision), 10)
+	}
+	return string(id)
+}
+
 // ParseTerm takes a termID as a string and parses it into a Term.
 // A complete term is in the form:
 // tenant:owner/name/revision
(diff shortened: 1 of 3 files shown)


meta.go
```diff
@@ -272,6 +274,36 @@ func (t *TermID) Validate() error {
 	return nil
 }
 
+// String returns the term in it's canonical string format.
```
rogpeppe: s/it's/its/

or even "in canonical form" ?

meta.go
```diff
@@ -272,6 +274,36 @@ func (t *TermID) Validate() error {
...
+		id = fmt.Sprintf("%s:", t.Tenant)
+	}
+	if t.Owner != "" && id == "" {
+		id = t.Owner
+	}
+	if id == "" {
+		id = t.Name
+	} else {
+		if t.Owner == "" {
+			id = fmt.Sprintf("%s%s", id, t.Name)
```
mhilton: If owner is blank don't we expect id to still be blank? does this line ever do anything? or is tenant:name a valid id. If it is can you list it above?

meta.go
```diff
@@ -272,6 +274,36 @@ func (t *TermID) Validate() error {
...
+		} else {
+			id = fmt.Sprintf("%s/%s", id, t.Name)
+		}
+	}
+	if t.Revision != 0 {
+		id = fmt.Sprintf("%s/%d", id, t.Revision)
+	}
+	return id
```
mhilton: I find this function quite confusing, how about
```go
func (t *TermsId) String() string {
	parts := make([]string, 0, 3)
	if t.Owner != "" {
		parts = append(parts, t.Owner)
	}
	parts = append(parts, t.Name)
	if t.Revision != 0 {
		parts = append(parts, fmt.Sprintf("%d", t.Revision))
	}
	id := strings.Join(parts, "/")
	if t.Tenant == "" {
		return id
	}
	return strings.Join([]string{t.Tenant, id}, ":")
}
```

meta.go
```diff
@@ -272,6 +274,36 @@ func (t *TermID) Validate() error {
...
+// This would be one of:
+//   tenant:owner/name/revision
+//   owner/name/revision
+//   owner/name
+//   name/revision
+//   name
+func (t *TermsId) String() string {
+	id := ""
+	if t.Tenant != "" {
+		id = fmt.Sprintf("%s:", t.Tenant)
```
rogpeppe: I find this a little hard to read, and it could be a bit more efficient too.

How about this, which is 5 lines shorter and quite a bit more efficient?

	func (t *TermsId) String() string {
		id := make([]byte, 0, len(t.Tenant)+1+len(t.Owner)+1+len(t.Name)+4)
		if t.Tenant != "" {
			id = append(id, t.Tenant...)
			id = append(id, ':')
		}
		if t.Owner != "" {
			id = append(id, t.Owner...)
			id = append(id, '/')
		}
		id = append(id, t.Name...)
		if t.Revision != 0 {
			id = append(id, '/')
			id = strconv.AppendInt(id, int64(t.Revision), 10)
		}
		return string(id)
	}

mhilton:
:+1: with some small thoughts

meta.go
```diff
@@ -291,10 +323,10 @@ func ParseTerm(s string) (*TermID, error) {
...
 
 	tokens := strings.Split(termid, "/")
-	var term TermID
+	var term TermsId
 	switch len(tokens) {
 	case 1: // "name"
-		term = TermID{
+		term = TermsId{
 			Tenant:   tenant,
 			Owner:    "",
```
rogpeppe: d

meta.go
```diff
@@ -303,14 +335,14 @@ func ParseTerm(s string) (*TermID, error) {
 	case 2: // owner/name or name/123
 		termRevision, err := strconv.Atoi(tokens[1])
 		if err != nil { // owner/name
-			term = TermID{
+			term = TermsId{
 				Tenant:   tenant,
 				Owner:    tokens[0],
 				Name:     tokens[1],
 				Revision: 0,
```
rogpeppe: d

meta.go
```diff
@@ -303,14 +335,14 @@ func ParseTerm(s string) (*TermID, error) {
...
 			}
 		} else { // name/123
-			term = TermID{
+			term = TermsId{
 				Tenant:   tenant,
 				Owner:    "",
```
rogpeppe: d

rogpeppe:
LGTM with a suggestion for the String function. I *think* my solution is preferable to mhilton's in terms of efficiency but YMMV.

meta.go
cmars: _tenant:_ is a valid prefix that can be added to any valid term ID without one.

meta.go
```diff
@@ -422,6 +450,7 @@ func parseMeta(m map[string]interface{}) (*Meta, error) {
 		}
 		meta.MinJujuVersion = minver
 	}
+	meta.Terms = parseStringList(m["terms"])
```
mattyw: I need to add a test for this line

rogpeppe:
LGTM, ta!

mattyw:
$$merge$$

PR #218, opened 2016-08-24
metrics: Implemented plan: optional | required in metrics.yaml

mattyw:


metrics.go
```diff
@@ -22,6 +22,8 @@ const (
 	// Supported metric types.
 	MetricTypeGauge    MetricType = "gauge"
 	MetricTypeAbsolute MetricType = "absolute"
+	PlanRequired       string     = "required"
```
rogpeppe: Should these have their own type?

Anyway, they shouldn't be in the "supported metric types" section unless they are actually metric types.

rogpeppe: Also, if they've not got their own type, then I'd omit the explicit "string" type.

metrics.go
```diff
@@ -65,7 +67,8 @@ type Metric struct {
 // Metrics contains the metrics declarations encoded in the metrics.yaml
 // file.
 type Metrics struct {
-	Metrics map[string]Metric
+	Metrics map[string]Metric `yaml:"metrics"`
+	Plan    string            `yaml:"plan"`
```
rogpeppe: If the only two possibilities are "required" and "optional", why not just have a "Required bool" field ?

rogpeppe:
LGTM with a few comments coming from a position of ignorance :)

metrics.go
cmars: Good point! Yes, I think something like:
```
plan:
  required: true|false
```
is just as clear and concise, but more open.

cmars:
See comment above... might consider `optional: true|false` rather than `required: true|false`, to align with relation metadata (see https://github.com/juju/charm/blob/v5/meta.go#L124).

metrics.go
rogpeppe: +1

metrics.go
```diff
@@ -58,14 +58,20 @@ func (m MetricType) validateValue(value string) error {
...
 type Metric struct {
-	Type        MetricType
-	Description string
+	Type        MetricType `yaml:"type"`
+	Description string     `yaml:"description"`
+}
+
+// Plan represents the plan section of metrics.yaml
+type Plan struct {
+	Optional bool `yaml:"optional"`
```
rogpeppe: Required bool `yaml:",omitempty"`

?

metrics.go
```diff
@@ -78,6 +84,9 @@ func ReadMetrics(r io.Reader) (*Metrics, error) {
 	if err := goyaml.Unmarshal(data, &metrics); err != nil {
 		return nil, err
 	}
+	if metrics.Plan == nil {
```
rogpeppe: Perhaps you can just delete this now.

metrics.go
```diff
@@ -112,3 +121,11 @@ func (m Metrics) ValidateMetric(name, value string) error {
 	}
 	return metric.Type.validateValue(value)
 }
+
+// PlanRequired returns true if these metrics require a plan.
```
rogpeppe: // PlanRequired reports whether these metrics require a plan.

? (standard Go phrasing)

metrics.go
```diff
@@ -58,14 +58,20 @@ func (m MetricType) validateValue(value string) error {
...
+type Plan struct {
+	Required bool `yaml:"required"`
 }
 
 // Metrics contains the metrics declarations encoded in the metrics.yaml
 // file.
 type Metrics struct {
-	Metrics map[string]Metric
+	Metrics map[string]Metric `yaml:"metrics"`
+	Plan    *Plan             `yaml:"plan,omitempty"`
```
cmars: If this was a value rather than a pointer, then you wouldn't need to check nil below -- it'd just have the default value when omitted.

metrics.go
```diff
@@ -112,3 +121,11 @@ func (m Metrics) ValidateMetric(name, value string) error {
 	}
 	return metric.Type.validateValue(value)
 }
+
+// PlanRequired returns true if these metrics require a plan.
+func (m Metrics) PlanRequired() bool {
+	if m.Plan == nil {
```
rogpeppe: return m.Plan != nil && m.Plan.Required

metrics.go
rogpeppe: The down side of that is it's not possible to omitempty a value struct.

rogpeppe:
LGTM modulo a few further suggestions, thanks!

metrics.go
cmars: Ah, good to know! It'll have to be a pointer then.

metrics.go
```diff
@@ -58,14 +58,20 @@ func (m MetricType) validateValue(value string) error {
...
 type Metric struct {
-	Type        MetricType
-	Description string
+	Type        MetricType `yaml:"type"`
+	Description string     `yaml:"description"`
+}
+
+// Plan represents the plan section of metrics.yaml
+type Plan struct {
+	Required bool `yaml:"required"`
```
rogpeppe: omitempty?

mattyw:
$$merge$$

PR #219, opened 2016-09-15
Update test to match changed juju/version behaviour

babbageclunk:
Since commit [4ae6172](https://github.com/juju/version/commit/4ae6172c00626779a5a462c3e3d22fc0e889431a) (https://github.com/juju/version/pull/3), `version.Number.Tag` will never include any digits, so this test became incorrect.


babbageclunk:
$$merge$$

PR #220, opened 2016-09-15
Remove charm.URL.Channel

babbageclunk:
It's no longer part of the charm URL, it's handled at a higher level.

(Preliminary task for fixing http://pad.lv/1584193.)


babbageclunk:
$$merge$$

PR #223, opened 2016-09-26
Introduce validation functions for URLs

kat-co:
Standardized validation of charm/bundle names and series. Now uses the juju/errors package and the corresponding NotValid error type. jujud will be calling these validation methods when charms are added to the environment to clean up error messages and fail faster.

This is to assist with fixing lp:1616200.


kat-co:
$$merge$$

PR #224, opened 2016-10-03
Revert charm url v3

kat-co:
This functionality is not in the charmstore and not yet consumed by Juju. Because the charmstore doesn't support v3 URLs yet, Juju cannot easily consume new versions of this library. For that reason, the v3 URL support is being forked and then reverted in this branch.


kat-co:
$$merge$$

PR #225, opened 2016-11-07
Update metrics to support plan:required, supporting LP:#1639855 fix.

cmars:
Also skipping old js mapreduce test, because it doesn't work on mongodb
without js engine, and this version of charm is no longer used in the
charmstore.


cmars:
$$merge$$

fabricematrat:
:+1:

PR #226, opened 2017-02-21
Add isCharm

bac:


charmdir.go
```diff
@@ -29,6 +29,17 @@ type CharmDir struct {
 // Trick to ensure *CharmDir implements the Charm interface.
 var _ Charm = (*CharmDir)(nil)
 
+// IsCharmDir returns true if the path reasonably represents a charm, even
```
rogpeppe: Standard go phrasing for bool-returning functions uses "reports where",
for example:

    // Is CharmDir reports whether the path is likely to represent
    // a charm, even if it cannot be opened as one.

charmdir.go
```diff
@@ -29,6 +29,17 @@ type CharmDir struct {
...
+// if it is later shown to be problematic.
+func IsCharmDir(path string) bool {
+	dir := &CharmDir{Path: path}
```
rogpeppe: I don't think I'd bother making the charm dir. Also, there's
no need to call os.Open (if you do, you need to remember to
close the reult).

I'd suggest:

      _, err := os.Stat(filepath.Join(dir, "metadata.yaml"))
      return err == nil

bac:
$$merge$$

PR #229, opened 2017-06-30
Add ActionSpec parameter to describe if an action is non-blocking.

ExternalReality:
Actions can now be specified as `blocking` the default or `non-blocking` with the default `blocking`. This is controlled by setting a property `NonBlocking: true`, for example, in an action's specification in a charm definition.

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -35,6 +35,7 @@ func NewActions() *Actions {
 // http://json-schema.org/draft-04/schema# (see http://json-schema.org/latest/json-schema-core.html)
 type ActionSpec struct {
 	Description string
+	NonBlocking bool
 	Params      map[string]interface{}
 }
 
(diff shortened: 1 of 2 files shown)


actions.go
```diff
@@ -141,6 +143,15 @@ func ReadActionsYaml(r io.Reader) (*Actions, error) {
 					return nil, errors.Errorf("value for schema key %q must be a string", key)
 				}
 				thisActionSchema[key] = typed
+			case "nonblocking":
```
ExternalReality: Notice that if key "nonblocking" is not found in the map it then defaults to false. Thus existing actions that do not specify the property `NonBlocking` will be counted as having `NonBlocking: false` which their existing behavior.

mitechie:
Should we have this be "blocking" vs "nonblocking"? It seems like potential trouble defining the thing to be a negative.

PR #231, opened 2017-09-01
Add git info to the zip file in local charm deployments.

Pekkari:
This pull request attempt to facilitate the debugging of local charm deployments
including the git info on the zip file, making it easy to spot if there is any local difference
in the charm in place, or allowing to push new changes from the local debugging environment.

Signed-off-by: José Pekkarinen <jose.pekkarinen@canonical.com>

### DIFF
--- a/charmdir.go
+++ b/charmdir.go
@@ -228,7 +228,7 @@ func (zp *zipPacker) visit(path string, fi os.FileInfo, err error) error {
 		if relpath == "build" {
 			return filepath.SkipDir
 		}
-		if hidden {
+		if hidden && !strings.HasPrefix(relpath, ".git") {
 			return filepath.SkipDir
 		}
 		relpath += "/"
@@ -242,7 +242,7 @@ func (zp *zipPacker) visit(path string, fi os.FileInfo, err error) error {
 	if mode&os.ModeSymlink != 0 {
 		method = zip.Store
 	}
-	if hidden || relpath == "revision" {
+	if hidden && !strings.HasPrefix(relpath, ".git") || relpath == "revision" {
 		return nil
 	}
 	h := &zip.FileHeader{


mitechie:
Can you fill us in a bit more on what you're looking to do? We intentionally don't include git info because it could be a private git repository, the data in there could be large and we don't want the size of the charm zip uploaded to the charm store to be unnecessarily large. The title here says "allow" but I'm missing how this is optional at this time?

Pekkari:
Well, it's not optional, perhaps I choose a bad title, sorry. 

I'm looking for different ways to improve the way juju operates on git repos if any. This is an easy way to get things hugely improved in the task of debugging environments deployed from internal git repositories, allowing to see which commit was deployed in which environment, and if it happen to be any local change. Deployments from git repos directly would be awesome, but meanwhile this may do a good trick for the team.

mitechie:
Rather than including the git directory, maybe it'd be better if something like charm build could detect a git repo and dump out the current git sha to a file in the charm?

rogpeppe:
I'm not sure that this is a good idea. I'm fairly sure that VCS information (including .git files) was one of the reasons we started skipping hidden files in the first place, because the VCS history can be huge and we don't really want that taking up space in the charm store or on deployed units.

ISTM that this change would result in the .git directories being included in charms uploaded to the charm store as well as in local deployments, and that doesn't sound like what you want here.

A straightforward way to attain what you want would be to run a little script before deploying the charm:

    git describe > git-info

Then you could look in that file to find out what's been deployed.

Pekkari:
no, that doesn't work, I want to get to a deployed environment and be able to see what is the commit I have deployed, and if there is any local changes, no matter when it was deployed. In other words, we want deterministic way to know what we deployed in the field without having any knowledge about when and how the deployment happened. I know git dir can be huge if your git repository is updated frequently but this is not our case.

mitechie:
I see, in that case then I'd suggest you have your charms pull directly from git as a deployment mechanism or option during deploy so that you have the full source tree. This is not something we can support out of the box on the charm command itself due to the issues we've mentioned here. (size and security mainly)

Pekkari:
@mitechie That can work, still misses the possibility of local changes.

PR #238, opened 2018-01-19
1640223 action name get function

AdrianGPrado:
[LP 1640223](https://bugs.launchpad.net/juju/+bug/1640223)

Variable `actionNameRule` is needed inside juju's repository
packages. By directly importing the variable in the juju package,
the code losses readability due to the fact that it doesn't
specify the origin of the variable.

With a function the intent of importing the variable from a different
scope is more clear.

  - [x] Code in `juju/juju/cmd/juju/action/run.go` without function `GetActionNameRule`:

          var keyRule = actionNameRule

    This passed the tests. But it doesn't specify what is actionNameRule nor where does it come from.

  - [x] Code in `juju/juju/cmd/juju/action/run.go` with function `GetActionNameRule`:

         import (
		    ...
            "gopkg.in/juju/charm.v6"
	        ...
         )

         var keyRule *regexp.Regexp = charm.GetActionNameRule()

   Here it explicitly imports `charm.v6` package and it specifies that `keyRule` will be a pointer to a regular expression. And it passes the tests.

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -19,7 +19,12 @@ var prohibitedSchemaKeys = map[string]bool{"$ref": true, "$schema": true}
 
 var actionNameRule = regexp.MustCompile("^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$")
 
-// Actions defines the available actions for the charm.  Additional params
+// Export `actionNameRule` variable to different contexts.
+func GetActionNameRule() *regexp.Regexp {
+	return actionNameRule
+}
+
+// Actions defines the available actions for the charm. Additional params
 // may be added as metadata at a future time (e.g. version.)
 type Actions struct {
 	ActionSpecs map[string]ActionSpec `yaml:"actions,omitempty" bson:",omitempty"`
--- a/actions_test.go
+++ b/actions_test.go
@@ -486,7 +486,7 @@ snapshot:
 					"properties":  map[string]interface{}{},
 				}}}},
 	}, {
-		description: "A simple snapshot actions YAML with numeric characters.",
+		description: "A simple snapshot actions YAML ending characters.",
 		yaml: `
 snapshot-01:
    description: Take database first snapshot.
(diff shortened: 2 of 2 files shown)


actions_test.go
```diff
@@ -508,6 +508,29 @@ snapshot-01:
 							"description": "The file to write out to.",
 							"type":        "string"}},
 					"required": []interface{}{"outfile"}}}}},
+	}, {
+		description: "A simple snapshot actions YAML starting characters.",
```
hatched: Possible copy/paste error? This looks to be identical to the above test.

actions_test.go
```diff
@@ -486,7 +486,7 @@ snapshot:
 					"properties":  map[string]interface{}{},
 				}}}},
 	}, {
-		description: "A simple snapshot actions YAML with numeric characters.",
+		description: "A simple snapshot actions YAML ending characters.",
```
hatched: I'm not sure about this edit? The description no longer makes sense....it is maybe 'with ending' ?

actions.go
```diff
@@ -19,7 +19,13 @@ var prohibitedSchemaKeys = map[string]bool{"$ref": true, "$schema": true}
 
 var actionNameRule = regexp.MustCompile("^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$")
 
-// Actions defines the available actions for the charm.  Additional params
+// Export `actionNameRule` variable to different contexts. The best practice
```
hatched: probably don't need to explain that it's a best practice in the code comments

actions_test.go
hatched: Sorry I now see that the label number switches from a suffix to a prefix.

hatched:
Should we maybe now also add a test for that exported function? Now that it'll be part of the public API.

actions.go
AdrianGPrado: Updated

actions_test.go
AdrianGPrado: Added `with names` so it's more clear.

AdrianGPrado:
$$merge$$

PR #241, opened 2018-05-17
Use set from the juju/collections package.

howbazaar:
As part of the work to break up juju/utils the set type has moved to juju/collections.

### DIFF
--- a/charmarchive.go
+++ b/charmarchive.go
@@ -14,7 +14,7 @@ import (
 	"path/filepath"
 	"strconv"
 
-	"github.com/juju/utils/set"
+	"github.com/juju/collections/set"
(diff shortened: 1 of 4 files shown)


mitechie:
:shipit:

mitechie:
doh, doesn't look like there's a bot on this. using the big green button and feels scared...

rogpeppe:
@mitechie FWIW the bot in this uses `$$merge$$` not `:shipit:` as it was originally part of the juju-core domain. Wouldn't it be nice to have the same merge instruction everywhere! If in doubt, look at the way that previous PRs have been merged.

mitechie:
@rogpeppe doh, thanks for the clarification

PR #245, opened 2018-06-18
added Device metadata

ycliuhw:
Added `Device` metadata for supporting `GPU` device on `CAAS`


meta.go
```diff
@@ -122,6 +122,44 @@ type Storage struct {
...
+	Name string `bson:"name"`
+
+	// Description is a description of the device.
+	Description string `bson:"description"`
+
+	// Type is the device type
+	Type DeviceType `bson:"type"`
+
+	// Request is the min number of devices to use
+	Request int64 `bson:"request"`
```
wallyworld: Instead of Request and Limit, we should stick to the convention used by Storage, CountMin and CountMax

meta.go
```diff
@@ -461,32 +504,34 @@ func (m Meta) MarshalYAML() (interface{}, error) {
...
+		Description   string                       `yaml:"description"`
+		Provides      map[string]marshaledRelation `yaml:"provides,omitempty"`
+		Requires      map[string]marshaledRelation `yaml:"requires,omitempty"`
+		Peers         map[string]marshaledRelation `yaml:"peers,omitempty"`
+		ExtraBindings map[string]interface{}       `yaml:"extra-bindings,omitempty"`
+		Categories    []string                     `yaml:"categories,omitempty"`
+		Tags          []string                     `yaml:"tags,omitempty"`
+		Subordinate   bool                         `yaml:"subordinate,omitempty"`
+		Series        []string                     `yaml:"series,omitempty"`
+		// Device         map[string]Device                `bson:"device,omitempty"`
```
wallyworld: Uncomment

meta.go
```diff
@@ -461,32 +504,34 @@ func (m Meta) MarshalYAML() (interface{}, error) {
...
+		Description:   m.Description,
+		Provides:      marshaledRelations(m.Provides),
+		Requires:      marshaledRelations(m.Requires),
+		Peers:         marshaledRelations(m.Peers),
+		ExtraBindings: marshaledExtraBindings(m.ExtraBindings),
+		Categories:    m.Categories,
+		Tags:          m.Tags,
+		Subordinate:   m.Subordinate,
+		Series:        m.Series,
+		// Device:         m.Device,
```
wallyworld: uncomment

meta.go
```diff
@@ -832,6 +897,37 @@ func parseStorage(stores interface{}) map[string]Storage {
...
+func parseDevice(devices interface{}) (map[string]Device, error) {
+	if devices == nil {
+		return nil, nil
+	}
+	result := make(map[string]Device)
+	for name, device := range devices.(map[string]interface{}) {
+		deviceMap := device.(map[string]interface{})
+		deviceType, ok := deviceMap["type"].(string)
+		if !ok {
+			return nil, fmt.Errorf("%q has invalid device type", name)
```
wallyworld: we should check that the value is valid also

meta.go
ycliuhw: I was trying to use k8s resource limit/request naming. But u r right, it should be consistent with Storage. updated

meta.go
ycliuhw: I am not sure where this MarshalYAML is used. And I can't find the `Storage` here. Should I remove here or uncomment?

meta.go
ycliuhw: yes, u r right. we have the schema check and we have a `Check` func after `parse`. 
The reason I do this check here is I wanna ensure `DeviceType(deviceType)` will not panic if `deviceMap["type"]` is nil or non string

meta.go
```diff
@@ -122,6 +122,44 @@ type Storage struct {
...
+func (dt *DeviceType) Validate() error {
+	for _, t := range DeviceTypes {
+		if *dt == t {
+			return nil
+		}
+	}
+	return fmt.Errorf("Invalid device type: %q, current supported are %+v", dt, DeviceTypes)
+}
+
+// Device represents a charm's device requirement(GPU for example).
```
wallyworld: space before the (

meta.go
```diff
@@ -122,6 +122,44 @@ type Storage struct {
...
+	// name of the GPU device on k8s could be like: nvidia.com/gpu
+	Name string `bson:"name"`
+
+	// Description is a description of the device.
+	Description string `bson:"description"`
+
+	// Type is the device type
+	Type DeviceType `bson:"type"`
+
+	// CountMin is the min number of devices to use
```
wallyworld: end comments with "."

meta.go
```diff
@@ -122,6 +122,44 @@ type Storage struct {
...
+			return nil
+		}
+	}
+	return fmt.Errorf("Invalid device type: %q, current supported are %+v", dt, DeviceTypes)
+}
+
+// Device represents a charm's device requirement(GPU for example).
+type Device struct {
+	// Name is the name of the device.
+	// name of the GPU device on k8s could be like: nvidia.com/gpu
```
wallyworld: This isn't quite right - see the spec. The name is a meaningful name to the charm.
The Type attribute below is where the k8s specific info goes.

meta.go
```diff
@@ -433,6 +472,10 @@ func parseMeta(m map[string]interface{}) (*Meta, error) {
 	}
 	meta.Series = parseStringList(m["series"])
 	meta.Storage = parseStorage(m["storage"])
+	meta.Device, err = parseDevice(m["device"])
```
wallyworld: Let's in-line the error check and also do the same for the extra bindings above as a drive by fix

meta.go
wallyworld: I am not sure either, but let's add storage as a drive by fix and make it complete

meta.go
```diff
@@ -650,6 +695,26 @@ func (meta Meta) Check() error {
...
 
+	for name, device := range meta.Device {
+		if device.Type == "" {
+			return fmt.Errorf("charm %q device %q: type must be specified", meta.Name, name)
+		}
+		if err := device.Type.Validate(); err != nil {
+			return err
+		}
+		if device.CountMin <= 0 {
+			return fmt.Errorf("charm %q device %q: invalid countmin amount %d", meta.Name, name, device.CountMin)
```
wallyworld: Let's do the error messages the same as for storage:

charm %q storage %q: invalid minimum count %d

Also 0 is valid as a min so the check is < 0

And as per what's done for storage we also need to check for duplicate names

meta.go
```diff
@@ -859,6 +955,22 @@ var storageSchema = schema.FieldMap(
 	},
 )
 
+var deviceSchema = schema.FieldMap(
+	schema.Fields{
+		"description": schema.String(),
+		"type": schema.OneOf(
+			schema.Const(string(DeviceGPU)),
+		),
+		"countmin": schema.Int(),
```
wallyworld: Let's copy off how storage does this using a schema (see storageCountC)
That means also that the parseDevices () method doesn't need to return an error as the checks are done by the schema

meta.go
```diff
@@ -929,6 +1041,7 @@ var charmSchema = schema.FieldMap(
 		"tags":             schema.List(schema.String()),
 		"series":           schema.List(schema.String()),
 		"storage":          schema.StringMap(storageSchema),
+		"device":           schema.StringMap(deviceSchema),
```
wallyworld: devices

meta_test.go
```diff
@@ -964,6 +964,102 @@ extra-bindings:
 	})
 }
 
+func (s *MetaSuite) TestDevice(c *gc.C) {
```
wallyworld: Devices

meta_test.go
```diff
@@ -964,6 +964,102 @@ extra-bindings:
...
+	meta, err := charm.ReadMeta(strings.NewReader(`
+name: a
+summary: b
+description: c
+device:
+    nvidia.com/gpu:
```
wallyworld: This is the type
The name is "bitcoin-miner" or something meaningful to the charm

meta.go
```diff
@@ -122,6 +122,40 @@ type Storage struct {
 	Properties []string `bson:"properties,omitempty"`
 }
 
+// DeviceType defines a device type.
+type DeviceType string
+
+// DeviceGPUGeneric represents generic GPU device type.
+const DeviceGPUGeneric DeviceType = "gpu"
```
wallyworld: Wrap these in a single const block
const (
  // comment
  blah

  // comment
  blah
)

wallyworld: We shouldn't encode this sort of validation here in the charm. The charm should just check that device type is not empty. Juju should then later validate the values based on the capabilities of the container substrate. eg we don't want to have to update this library when k8s adds a new device type

meta.go
```diff
@@ -122,6 +122,40 @@ type Storage struct {
...
+	Description string `bson:"description"`
+
+	// Type is the device type.
+	// currently supported types are
+	// - gpu
+	// - nvidia.com/gpu
+	// - amd.com/gpu
+	Type DeviceType `bson:"type"`
+
+	// CountMin is the min number of devices to use.
```
wallyworld: ... of devices that the charm requires.

ycliuhw:
$$merge$$

PR #246, opened 2018-06-21
Charm version file create/overwrite

vinu2003:
This PR has helper function for creating or overwriting charmn version file.


vinu2003:
Working on unit test cases.

charm.go
```diff
@@ -99,3 +101,52 @@ func IsUnsupportedSeriesError(err error) bool {
 	_, ok := err.(*unsupportedSeriesError)
 	return ok
 }
+
+// MaybeCreateVersionFile creates/overwrite charm version file.
+func MaybeCreateVersionFile(path string) error {
+	var charmVersion string
```
wallyworld: Don't declare charmVersion here - it is used way down below.
Variables need to be defined/initialised in the block where they're used.

charm.go
```diff
@@ -99,3 +101,52 @@ func IsUnsupportedSeriesError(err error) bool {
...
+	var cmdArgs []string
+	var err error
+	// Verify that it is revision control directory.
+	if _, err = os.Stat(filepath.Join(path, ".git")); err == nil {
```
wallyworld: Use err :=
Then no need for err var above

charm.go
```diff
@@ -99,3 +101,52 @@ func IsUnsupportedSeriesError(err error) bool {
...
+	var args []string
+	for pos, arg := range cmdArgs {
+		if pos != 0 {
+			args = append(args, arg)
+		}
+	}
+	cmd := exec.Command(cmdArgs[0], args...)
+	outStr, err := cmd.CombinedOutput()
+	if err != nil {
+		logger.Infof("Command output: %v", outStr)
```
wallyworld: This will be displayed to the user - don't log at info level.
If we do log at debug level, it needs more context, eg "charm revision info:"

charm.go
```diff
@@ -99,3 +101,52 @@ func IsUnsupportedSeriesError(err error) bool {
...
+		return err
+	}
+	charmVersion = string(outStr)
+
+	versionPath := filepath.Join(path, "version")
+	// Overwrite the existing version file.
+	file, err := os.OpenFile(versionPath, os.O_RDWR|os.O_CREATE, 0666)
```
wallyworld: 0644

charm.go
```diff
@@ -99,3 +101,44 @@ func IsUnsupportedSeriesError(err error) bool {
...
+		// It is baazar.
+		cmdArgs = []string{"bzr", "revision-info"}
+	} else if _, err = os.Stat(filepath.Join(path, ".hg")); err == nil {
+		cmdArgs = []string{"hg", "id", "--id"}
+	} else {
+		logger.Debugf("Charm is not in revision control directory")
+		return nil
+	}
+
+	var charmVersion string
```
wallyworld: no need for var here, just assign down below

charm_test.go
```diff
@@ -219,3 +241,98 @@ func cloneDir(c *gc.C, path string) string {
...
+	_, err := os.Create(gitPath)
+	c.Assert(err, gc.IsNil)
+
+	err = charm.MaybeCreateVersionFile(tempPath)
+	c.Assert(err, gc.IsNil)
+
+	testing.AssertEchoArgs(c, "git", expectedArgs.gitBase...)
+
+	versionPath := filepath.Join(tempPath, "version")
+	_, err = os.Stat(versionPath)
```
wallyworld: We want to read the contents of the version path file and check that the contents are as expected

charm_test.go
```diff
@@ -12,19 +12,41 @@ import (
...
 func Test(t *stdtesting.T) {
 	gc.TestingT(t)
 }
 
-type CharmSuite struct{}
+type CharmSuite struct {
+	coretesting.BaseSuite
+}
+
+var expectedArgs = struct {
```
wallyworld: don't need this struct as there are separate tests.
just use the args slices directly in the relevant tests

charm_test.go
vinu2003: done

charm_test.go
vinu2003: done

charm_test.go
```diff
@@ -219,3 +223,132 @@ func cloneDir(c *gc.C, path string) string {
 	c.Assert(err, gc.IsNil)
 	return newPath
 }
+
+// TestCreateMaybeCreateVersionFile verifies if the version file can be created
+// in case of git revision control directory
+func (s *CharmSuite) TestGitMaybeCreateVersionFile(c *gc.C) {
```
wallyworld: given how much code is common in all these 3 functions, extract a help method called assertVersionFile(c *gc.C, dir, execName string, args ...string)
 Then the logic uses these params.
Each of the 3 test methods then become something like

func (s *CharmSuite) TestGitMaybeCreateVersionFile(c *gc.C) {
    s.assertVersionFile(c, ".git", "git", "describe", "--dirty")
}

charm.go
```diff
@@ -99,3 +101,41 @@ func IsUnsupportedSeriesError(err error) bool {
 	_, ok := err.(*unsupportedSeriesError)
 	return ok
 }
+
+// MaybeCreateVersionFile creates/overwrite charm version file.
+func MaybeCreateVersionFile(path string) error {
+	var cmdArgs []string
+	// Verify that it is revision control directory.
+	if _, err := os.Stat(filepath.Join(path, ".git")); err == nil {
+		// It is git version control.
```
wallyworld: We don't need these comments, it's obvious

charm.go
```diff
@@ -99,3 +101,41 @@ func IsUnsupportedSeriesError(err error) bool {
...
+		cmdArgs = []string{"git", "describe", "--dirty"}
+	} else if _, err = os.Stat(filepath.Join(path, ".bzr")); err == nil {
+		// It is baazar.
+		cmdArgs = []string{"bzr", "revision-info"}
+	} else if _, err = os.Stat(filepath.Join(path, ".hg")); err == nil {
+		cmdArgs = []string{"hg", "id", "--id"}
+	} else {
+		logger.Debugf("Charm is not in revision control directory")
```
wallyworld: generally start debug strings with lower caae

vinu2003:
!!build!!

vinu2003:
!!build!!

charm.go
```diff
@@ -99,3 +101,39 @@ func IsUnsupportedSeriesError(err error) bool {
...
+		cmdArgs = []string{"git", "describe", "--dirty"}
+	} else {
+		logger.Debugf("Charm is not in revision control directory")
+		return nil
+	}
+
+	outStr, err := utils.RunCommand(cmdArgs[0], cmdArgs[1:]...)
+	if err != nil {
+		return err
+	} else {
```
wallyworld: Don't need the else{} bit. It just introduces unnecessary indenting.

charm.go
```diff
@@ -99,3 +101,39 @@ func IsUnsupportedSeriesError(err error) bool {
...
+		versionPath := filepath.Join(path, "version")
+		// Overwrite the existing version file.
+		file, err := os.OpenFile(versionPath, os.O_RDWR|os.O_CREATE, 0644)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+
+		_, err = file.WriteString(outStr)
```
wallyworld: We generally inline things like this as:

if _, err = file.WriteString(outStr); err != nil {
    return err
}

charm_test.go
```diff
@@ -4,27 +4,33 @@
...
 
+	"github.com/juju/testing"
 	jc "github.com/juju/testing/checkers"
 	"github.com/juju/utils/fs"
 	gc "gopkg.in/check.v1"
+	"gopkg.in/juju/charm.v6"
 	"gopkg.in/yaml.v2"
 
-	"gopkg.in/juju/charm.v6"
+	coretesting "github.com/juju/juju/testing"
```
wallyworld: This charm.v6 repo must not depend on juju/juju

charm_test.go
```diff
@@ -4,27 +4,33 @@
...
 )
 
 func Test(t *stdtesting.T) {
 	gc.TestingT(t)
 }
 
-type CharmSuite struct{}
+type CharmSuite struct {
+	coretesting.BaseSuite
```
wallyworld: remove this BaseSuite - we only use it in juju/juju

charm_test.go
```diff
@@ -219,3 +225,88 @@ func cloneDir(c *gc.C, path string) string {
 	c.Assert(err, gc.IsNil)
 	return newPath
 }
+
+func (s *CharmSuite) assertVersionFile(c *gc.C, dir string, execName string, args []string) {
```
wallyworld: Can we rename dir to charmDir so it's clear what dir it is
Actually, we are always passing in the same charm path for "dummy"  so no need to make this a param. What I was meaning in my original comment was to pass in the vcs dir, ".git" etc. But since it seems we can construct that always from the execName, then there's no need to pass it in at all. So this dir param can go.

charm_test.go
```diff
@@ -219,3 +225,88 @@ func cloneDir(c *gc.C, path string) string {
...
+	testing.PatchExecutableAsEchoArgs(c, s, execName)
+
+	// copy all the contents from 'path' to 'tmp folder dummy-charm'
+	// Using cloneDir
+	tempPath := cloneDir(c, dir)
+
+	// create an empty .execName file inside tempDir
+	lookupDir := "." + execName
```
wallyworld: No need for explicit var, just do
vcsPath := filepath.Join(tempPath, "." + execName)

Some people would also even do

_, err := os.Create(filepath.Join(tempPath, "." + execName))

charm_test.go
```diff
@@ -219,3 +225,88 @@ func cloneDir(c *gc.C, path string) string {
...
+	expectedVersion := make([]string, 1, 2)
+	for pos := range args {
+		args[pos] = "'" + args[pos] + "'"
+	}
+	expectedVersion[0] = execName + " " + strings.Join(args, " ")
+
+	f, err := os.Open(versionPath)
+	c.Assert(err, gc.IsNil)
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
```
wallyworld: No need for scanner.
Just do ioutils.ReadAll()
That will get the file content which should just be one line of text.

charm_test.go
```diff
@@ -219,3 +225,88 @@ func cloneDir(c *gc.C, path string) string {
...
+
+	testing.AssertEchoArgs(c, execName, args...)
+
+	// Verify if version exists.
+	versionPath := filepath.Join(tempPath, "version")
+	_, err = os.Stat(versionPath)
+	c.Assert(err, gc.IsNil)
+
+	// Read the contents of version file.
+	var version []string
```
wallyworld: Don't declare this all the way up here when it's not used till down below. Try and declare variables close to where they're used

charm_test.go
```diff
@@ -219,3 +225,88 @@ func cloneDir(c *gc.C, path string) string {
...
+	expectedVersion[0] = execName + " " + strings.Join(args, " ")
+
+	f, err := os.Open(versionPath)
+	c.Assert(err, gc.IsNil)
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		version = append(version, scanner.Text())
+	}
+	c.Assert(version, gc.DeepEquals, expectedVersion)
```
wallyworld: We should just compare strings here so no need for DeepEquals

charm_test.go
```diff
@@ -219,3 +225,88 @@ func cloneDir(c *gc.C, path string) string {
...
+	// Verify if version exists.
+	versionPath := filepath.Join(tempPath, "version")
+	_, err = os.Stat(versionPath)
+	c.Assert(err, gc.IsNil)
+
+	// Read the contents of version file.
+	var version []string
+
+	// Build from input arguments.
+	expectedVersion := make([]string, 1, 2)
```
wallyworld: We can write this as

expectedVersion := execName + " " + testing.ReadEchoArgs(c, execName)

charm.go
vinu2003: done

charm.go
vinu2003: done

charm_test.go
vinu2003: Ok. Iam using IsolationSuite from github.com/juju/testing

charm_test.go
vinu2003: Using IsolationSuite

charm_test.go
vinu2003: done

charm_test.go
vinu2003: Done

charm_test.go
vinu2003: done

charm_test.go
vinu2003: ReadArgs doesnt the work the way. I again gave few tries.happy to discuss if there another optimised way.

vinu2003:
!!build!!

charm_test.go
```diff
@@ -12,19 +12,22 @@ import (
...
 	stdtesting "testing"
 
+	"github.com/juju/testing"
 	jc "github.com/juju/testing/checkers"
 	"github.com/juju/utils/fs"
 	gc "gopkg.in/check.v1"
-	"gopkg.in/yaml.v2"
-
 	"gopkg.in/juju/charm.v6"
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
vinu2003: done

vinu2003:
!!build!!

charm_test.go
```diff
@@ -219,3 +223,83 @@ func cloneDir(c *gc.C, path string) string {
...
+	for _, arg := range args {
+		expectedVersion = fmt.Sprintf("%s %s", expectedVersion, util.ShQuote(arg))
+	}
+
+	f, err := os.Open(versionPath)
+	c.Assert(err, jc.ErrorIsNil)
+	defer f.Close()
+
+	var version []byte
+	version, err = ioutil.ReadAll(f)
```
wallyworld: version, err :=

vinu2003:
$$merge$$

PR #252, opened 2018-07-15
Only check file details for the filetype.

Veebers:
When dealing with Docker resource types we don't need to validate the file details.


Veebers:
$$merge$$

PR #255, opened 2018-08-03
Rename docker to oci-image for resources.

Veebers:
Renames the resource name that is exposed to charm authors from docker to oci-image.

### DIFF
--- a/resource/type.go
+++ b/resource/type.go
@@ -11,12 +11,12 @@ import (
 const (
 	typeUnknown Type = iota
 	TypeFile
-	TypeDocker
+	TypeContainerImage
 )
 
 var types = map[Type]string{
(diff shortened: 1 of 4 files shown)


resource/type.go
```diff
@@ -15,8 +15,8 @@ const (
 )
 
 var types = map[Type]string{
-	TypeFile:  "file",
-	TypeDocker: "docker",
+	TypeFile:   "file",
+	TypeDocker: "oci-image",
```
wallyworld: Let's make this TypeContainerImage (the juju use of docker can stay as is)

Veebers:
$$merge$$

Veebers:
$$merge$$

Veebers:
$$merge$$

PR #256, opened 2018-08-22
Ensure paths in charm archives use /

wallyworld:
When making a charm (zip) archive from a local charm directory, any relative paths on Windows contain the Windows path separator. Convert to "/" before adding to the archive as per zip file spec 4.4.17.1


wallyworld:
$$merge$$

PR #258, opened 2018-08-29
Improve charm version string (on git)

wallyworld:
The generation of the charm version string was failing on git if there were no tags or current branch.
Use the --always arg to fix this.

Also, improve the log warning to explain what happened if there is  an error (to avoid end user confusion).

And improve the tests.

### DIFF
--- a/charmdir.go
+++ b/charmdir.go
@@ -9,12 +9,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"syscall"
-
-	"github.com/juju/utils"
 )
 
 // The CharmDir type encapsulates access to data and operations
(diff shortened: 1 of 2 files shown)


charmdir.go
```diff
@@ -174,9 +173,13 @@ func resolveSymlinkedRoot(rootPath string) (string, error) {
...
 // By convention a charm archive should have a ".charm" suffix.
 func (dir *CharmDir) ArchiveTo(w io.Writer) error {
-	versionString, err := dir.MaybeGenerateVersionString()
+	versionString, vcsType, err := dir.MaybeGenerateVersionString()
 	if err != nil {
-		logger.Warningf("version string generation failed : %v", err)
+		// Just to be safe, ensure version is "" on error.
+		versionString = ""
+		logger.Warningf(
+			"%s version string generation failed : %v\nThis means that the charm version won't show in juju status.",
```
anastasiamac: I wonder if you want to have %q here instead... reading through https://golang.org/pkg/fmt/, %q seems better

charmdir.go
```diff
@@ -331,20 +334,33 @@ func checkFileType(path string, mode os.FileMode) error {
 }
 
 // MaybeGenerateVersionString generates charm version string.
-func (dir *CharmDir) MaybeGenerateVersionString() (string, error) {
+func (dir *CharmDir) MaybeGenerateVersionString() (string, string, error) {
```
anastasiamac: Maybe worth a comment what a 2nd returned string is?

wallyworld:
!!build!!

wallyworld:
$$merge$$

charmdir.go
wallyworld: ok

charmdir.go
wallyworld: I don't mind %s as the message reads fine with it. Changed to %q though in case it helps.

PR #259, opened 2018-08-29
Don't ignore an error

wallyworld:
https://github.com/juju/charm/pull/258 had some code which didn't check for an error from cmd.Output()


wallyworld:
$$merge$$

PR #261, opened 2018-08-31
charm: don't exclude all dot files.

### LINKED ISSUE #201: Stop removing all hidden files
currently, charm strips all hidden files from a charm payload but this is removing valuable information that's not related to VCS.

This strips files, like .build.manifest, which are used to record the layers in the built charm artifact.

mhemmings:
We now only exclude vcs directories, and it should be relatively easy to allow user defined ignores in the future too.

Closes #201.


charmdir.go
```diff
@@ -258,7 +268,7 @@ func (zp *zipPacker) visit(path string, fi os.FileInfo, err error) error {
 	if mode&os.ModeSymlink != 0 {
 		method = zip.Store
 	}
-	if hidden || relpath == "revision" {
+	if relpath == "revision" {
```
rogpeppe: tbh I it might be worth leaving the "hidden || " part there, for the moment if/when we allow excluding files as well was directories. If we have a file name ".git" I think it's reasonable to exclude that too.

Then shouldIgnoreDir becomes just shouldIgnore and the logic is almost the same as before, and it's pretty trivial to add a .charmignore file later if we want to.

rogpeppe:
$$merge$$

PR #263, opened 2018-09-06
WIP: add lxdprofile to charms

hmlanigan:
Add lxdprofile to a charm per: 
https://discourse.jujucharms.com/t/wip-specification-for-lxd-profile-updates-permitted-by-charms/78

Current tests are passing, lxdprofile_test.go has been added.  To be done: enhance current tests.

### DIFF
--- a/lxdprofile.go
+++ b/lxdprofile.go
@@ -46,10 +46,8 @@ func (profile *LXDProfile) HasName() bool {
 // ValidateConfigDevices validates the Config and Devices properties of the LXDProfile.
 // WhiteList devices: unix-char, unix-block, gpu, usb.
 // BlackList config: boot*, limits* and migration*.
+// An empty profile will not return an error.
 func (profile *LXDProfile) ValidateConfigDevices() error {
-	if len(profile.Devices) < 1 && len(profile.Config) < 1 {
-		return fmt.Errorf("invalid lxd-profile.yaml: does not contain devices nor config")
-	}
 	for _, val := range profile.Devices {
 		goodDevs := set.NewStrings("unix-char", "unix-block", "gpu", "usb")
 		if devType, ok := val["type"]; ok {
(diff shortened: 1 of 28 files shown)


hmlanigan:
This will be a little cleaner once https://github.com/juju/charm/pull/262 lands

lxdprofile.go
```diff
@@ -0,0 +1,69 @@
...
+// HasName returns true if the lxd profile name is set
+func (profile *LXDProfile) HasName() bool {
+	return profile.Name != ""
+}
+
+// ValidateConfigDevices validates the Config and Devices properties of the LXDProfile.
+// WhiteList devices: unix-char, unix-block, gpu, usb.
+// BlackList config: boot*, limits* and migration*.
+func (profile *LXDProfile) ValidateConfigDevices() error {
+	if len(profile.Devices) < 1 && len(profile.Config) < 1 {
```
SimonRichardson: It would be nice if we could ask if the profile is empty before validating the LXDProfile. The way the parsing of the file is done, the LXDProfile will always return a non-nil ptr (inline with other types inside charm code). So it would nice to either have `charm.IsLXDProfileEmpty()` or `lxdProfile.Empty()`, which can tell us if any items within the profile have been defined or not.

lxdprofile.go
```diff
@@ -0,0 +1,69 @@
...
+	"strings"
+
+	"github.com/juju/errors"
+	"github.com/juju/utils/set"
+	lxdapi "github.com/lxc/lxd/shared/api"
+	"gopkg.in/yaml.v2"
+)
+
+type LXDProfile struct {
+	lxdapi.Profile
```
SimonRichardson: I'm still unsure about this, I wonder if we should copy the structs over so that we can adapt if required in the future. 

Note: the lxdapi.Profile is made of multiple composed structs rather than just one. It might be nice to have one, so it's easier to reason about.

hmlanigan: after conversation agreed on using equivalent of lxdapi.ProfilePut.

lxdprofile.go
hmlanigan: adding lxdProfile.Empty() to keep in line with the other pieces of charm.

PR #264, opened 2018-09-12
WIP: add lxdprofilecharm interface

hmlanigan:
Adding for the lxd profile work in juju.  Charms can now read and validate an included lxd-profile.yaml.

### DIFF
--- a/lxdprofile.go
+++ b/lxdprofile.go
@@ -1,0 +1,74 @@
+// Copyright 2018 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package charm
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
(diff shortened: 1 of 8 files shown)


lxdprofile.go
```diff
@@ -0,0 +1,76 @@
...
+	"io/ioutil"
+	"strings"
+
+	"github.com/juju/errors"
+	"github.com/juju/utils/set"
+	"gopkg.in/yaml.v2"
+)
+
+type LXDProfileCharm interface {
+	Charm
```
SimonRichardson: I don't think we should make the `LXDProfileCharm` a composition of `Charm` and `LXDProfiler`. With that I think this should be renamed to `LXDProfiler`.

hmlanigan:
$$merge$$

PR #268, opened 2018-12-05
Don't indicate that all series are supported by old style charms.

ExternalReality:
The `SeriesForCharm` function would simply return the requested series when no series were supported by the charm. This leads errors in `juju`. No usage of `SeriesForCharm` relies on the suspect behavior.

### DIFF
--- a/charm.go
+++ b/charm.go
@@ -42,16 +42,16 @@ func ReadCharm(path string) (charm Charm, err error) {
 }
 
 // SeriesForCharm takes a requested series and a list of series supported by a
-// charm and returns the series which is relevant.
-// If the requested series is empty, then the first supported series is used,
-// otherwise the requested series is validated against the supported series.
+// charm and returns the series which is relevant. If the requested series is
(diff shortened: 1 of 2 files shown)


ExternalReality:
[test multi-series subordinate with upgrading series on machine and what does set-series do for subordinates?](https://trello.com/c/ltxZ5PBS/785-test-multi-series-subordinate-with-upgrading-series-on-machine-and-what-does-set-series-do-for-subordinates)

ExternalReality:
[Don't indicate that any series is supported by old style charms.](https://trello.com/c/8uK5BIrV/878-dont-indicate-that-any-series-is-supported-by-old-style-charms)

jameinel:
I haven't gone over everything with a fine tooth comb, but I definitely support the direction this is going in. We've had multi series support for long enough that people can list the series their charm really does support.

ExternalReality:
$$MERGE$$

ExternalReality:
I keep forgetting to ask @howbazaar ot look at this. Trust @jameinel's word merging...

PR #269, opened 2019-02-06
Expose charm version on archive

achilleasa:
This PR is part of the work required for fixing https://bugs.launchpad.net/juju/+bug/1812227. Juju needs to retrieve the VCS version (e.g git sha) for charm archives fetched from the charmstore. 

In order to avoid changing the `Charm` interface that would inevitably break other dependencies, the PR exposes a `Version()` method on the `CharmArchive` type. The actual juju code that fetches the archive (using the charmrepo pkg) already casts the result to a `charm.CharmArchive` so this change allows us to fetch the version with minimal impact to other users of the package.


achilleasa:
$$merge$$

PR #270, opened 2019-02-08
Revert "Don't indicate that all series are supported by old style charms."

achilleasa:
Reverts juju/charm#268

The changes introduced by the above PR break several juju tests in the 2.5 branch and especially the following ones that assume that this charm handles series selection in a particular way:
- https://github.com/juju/juju/blob/2.5/cmd/juju/application/series_selector_test.go#L16
- https://github.com/juju/juju/blob/2.5/cmd/juju/application/deploy_test.go#L720


achilleasa:
$$merge$$

PR #271, opened 2019-02-18
Jujuignore support for charm archiver

achilleasa:
This PR introduces support for `.jujuignore` files and refactors the default path exclusion list (used to contain VCS-related paths) to use the same rule syntax. The original default exclusion list has been refactored to include some additional patterns (`build/`, `revision` and a missing rule for `version`) which were enforced by `if relpath == X` predicates in the archiver code.

The charm archiver looks for a `.jujuignore` file at the root of the charm directory and merges its contents with the default path exclusion list. The set of exclusion rules is used to decide which files should be added to the charm archive. 

`.jujuignore` files use the same syntax as [gitignore](https://git-scm.com/docs/gitignore#_pattern_format). The pattern rules are summarized in the extended commit log for https://github.com/juju/charm/commit/4b65164cf6c488e9d1138494bed1676b6fa7cda8

The PR is part of the work required for fixing https://bugs.launchpad.net/juju/+bug/1813799


charmdir.go
```diff
@@ -16,26 +16,38 @@ import (
...
-}
+// defaultJujuIgnore contains jujuignore directives for excluding VCS- and
+// build-related directories when archiving. The following set of directives
+// will be appended to the contents of the charm's .jujuignore file if one is
+// provided.
+//
+// NOTE: writeArchive auto-generates its own revision and version files so they
+// need to be excluded here to prevent anyone from overriding their contents by
+// adding files with the same name to their charm repo.
+var defaultJujuIgnore = `
```
timClicks: From the bug report, it looks like tox is being noisy. Should we add `.tox` (and maybe `.tmp` as [it looks like that's used also](https://tox.readthedocs.io/en/latest/config.html#tox-global-settings))?

achilleasa: The idea is that we will be updating https://github.com/juju-solutions/template-reactive-python to include a `.jujuignore` file that excludes `.tox` so that should take care of that.

achilleasa: I could add all these bits to the default list but I am concerned it would probably defeat the purpose of having a jujuignore file in the first place.

charmdir.go
```diff
@@ -123,6 +139,33 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
...
+			return errors.Annotate(err, ".jujuignore")
+		}
+	}
+
+	defaultRules, err := newIgnoreRuleset(strings.NewReader(defaultJujuIgnore))
+	if err != nil {
+		return err
+	}
+
+	dir.ignoreRules = append(dir.ignoreRules, defaultRules...)
```
SimonRichardson: Is this right, where `defaultRules...` will overwrite the charm `ignoreRules`?

So if you have `foo` in default `.jujuignore` and then `!foo` in charm `.jujuignore` then it won't be included. Even if that's not possible, we should probably be consistent and add `defaultRules` first to be consistent?

achilleasa: You have a point there. Makes more sense to do it the other way round. 

An alternative we may want to consider is to use the charm's .jujuignore file if one is provided OR use the defaults as a fallback. This way there will be no unexpected surprises for people writing .jujuignore files. Thoughts?

SimonRichardson: I think it should be safe to just flip it the other way around (defaults first), mainly because I think it should be perfectly safe for them to do `!.git` if they require. Whilst also providing some very safe defaults for most users and it will help with cutting down on files we have to manage in general, as apposed to not providing one.

SimonRichardson: Also `.git` folder is entirely dangerous given that secrets accidentally committed and then removed from git history could still be extracted. So having sane defaults seems like a pure win.

PR #272, opened 2019-02-19
Switch to lazy generation for jujuignore rules

achilleasa:
The previous implementation stored the set of rules (a slice of
functions) into the charm struct. This proved problematic for tests that
perform deep-equal comparisons of charm structs (in Go, functions are
not comparable for equality).

Given that the ignore rules are only used when archiving, there is no
particular reason to keep them within the charm struct.


achilleasa:
$$merge$$

PR #274, opened 2019-04-26
New metadata for charm deployment info

wallyworld:
A new optional metadata block is added to charm metadata.yaml to allow charms to specify extra info about their deployment. This is added to allow k8s charms to say if they want a stateful or stateless controller, or what service type is preferred by default, eg

```
deployment:
    type: stateless
    service: loadbalancer
```

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 
 	"github.com/juju/errors"
+	"github.com/juju/os"
+	"github.com/juju/os/series"
 	"github.com/juju/schema"
 	"github.com/juju/utils"
 	"github.com/juju/version"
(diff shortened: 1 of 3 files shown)


meta.go
```diff
@@ -905,6 +943,47 @@ func parseDevices(devices interface{}) map[string]Device {
...
 }
 
+func parseDeployment(deployment interface{}, charmSeries []string, storage map[string]Storage) (*Deployment, error) {
+	if deployment == nil {
+		return nil, nil
+	}
+	if len(charmSeries) == 0 {
+		return nil, errors.New("charm with deployment metadata must declare at least one series")
+	}
+	deploymentMap := deployment.(map[string]interface{})
```
ycliuhw: !ok return error?

meta.go
```diff
@@ -147,6 +149,34 @@ type Device struct {
...
+	DeploymentStateful  DeploymentType = "stateful"
+)
+
+// ServiceType defines a service type.
+type ServiceType string
+
+const (
+	ServiceCluster      ServiceType = "cluster"
+	ServiceLoadBalancer ServiceType = "loadbalancer"
+	ServiceExternal     ServiceType = "external"
```
ycliuhw: just wondering how we are going to use `ServiceLoadBalancer` and `ServiceExternal`

wallyworld: We just create the k8s service with what the charm specifies.
It's up to the user to have their cluster correctly configured. We can figure out if there's any way to do some pre-checks of the cluster at deploy time, but that doesn't affect the metadata definition.

meta.go
wallyworld: no need, that validation is done by the schema definition

ycliuhw: ah, ic, thanks

wallyworld:
$$merge$$

PR #276, opened 2019-05-30
Add "trust" field to application/service spec

achilleasa:
This field is part of the bundle's yaml document and indicates that an
application requires access to cloud credentials and must be explicitly
trusted by the operator.


achilleasa:
$$merge$$

PR #277, opened 2019-06-14
Add "offers" block to application specification block

achilleasa:
This PR extends the application spec to include the list of offers exposed by the model. The offers are encoded as a map where the key is the offer name and values are OfferSpec instances. Currently, OfferSpecs only contain the endpoint name but in the future we could extend them with additional bits of information such as an ACL for the users who can interact (read, admin, consume) with the offer.


achilleasa:
$$merge$$

PR #280, opened 2019-06-25
Bundle data with SAAS

SimonRichardson:
The following adds a SAAS spec to the bundle data so that we can
start consuming the right information.

```yaml
series: bionic
saas:
    apache2:
        source: local
        url: production:admin/info.apache2
```

Drive by: removed trailing white spaces.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -176,8 +176,7 @@ func (d *BundleData) UnmarshaledWithServices() bool {
 // SaasSpec represents a single software as a service (SAAS) node.
 // This will be mapped to consuming of offers from a bundle deployment.
 type SaasSpec struct {
-	Source string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
-	URL    string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+	URL string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 }
 
 // MachineSpec represents a notional machine that will be mapped
@@ -547,10 +546,6 @@ func (verifier *bundleDataVerifier) verifySaas() {
 		}
 		if saas == nil {
 			continue
-		}
-		if saas.Source != "" && !names.IsValidControllerName(saas.Source) {
-			// source here could be "local" or a controller name
-			verifier.addErrorf("invalid source %q for SAAS %s", saas.Source, name)
 		}
 		if saas.URL != "" && !IsValidOfferURL(saas.URL) {
 			verifier.addErrorf("invalid offer URL %q for SAAS %s", saas.URL, name)
--- a/bundledata.go
+++ b/bundledata.go
@@ -136,6 +136,11 @@ type BundleData struct {
 	// not referred to by a unit placement directive.
 	Machines map[string]*MachineSpec `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 
+	// Saas holds one entry for each software as a service (SAAS) for cross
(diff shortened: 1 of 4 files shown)


SimonRichardson:
https://trello.com/c/yzlIu5DE/1524-cmr-update-charmv6-bundle-definition-for-the-consume-yaml-bits

bundledata.go
```diff
@@ -168,6 +173,13 @@ func (d *BundleData) UnmarshaledWithServices() bool {
 	return d.unmarshaledWithServices
 }
 
+// SaasSpec represents a single software as a service (SAAS) node.
+// This will be mapped to consuming of offers from a bundle deployment.
+type SaasSpec struct {
+	Source string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+	URL    string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+}
```
achilleasa: I wonder whether we need to add the controller UUID here so the client can avoid any confusion caused by similarly named controllers from its local cache.

@mitechie thoughts?

bundledata.go
```diff
@@ -168,6 +173,13 @@ func (d *BundleData) UnmarshaledWithServices() bool {
 	return d.unmarshaledWithServices
 }
 
+// SaasSpec represents a single software as a service (SAAS) node.
+// This will be mapped to consuming of offers from a bundle deployment.
+type SaasSpec struct {
+	Source string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
```
achilleasa: What is the purpose of `Source`? I am not sure whether we need to distinguish between local/remote controller offers; when you `juju consume` the client provides a macaroon and the controller details for the one exposing the offer so it shouldn't make any difference (I think).

offerurl.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// HasEndpoint returns whether this offer URL includes an
+// endpoint name in the application name.
+func (u *OfferURL) HasEndpoint() bool {
+	return strings.Contains(u.ApplicationName, ":")
+}
+
+// modelApplicationRegexp parses urls of the form controller:user/model.application[:relname]
+var modelApplicationRegexp = regexp.MustCompile(`(/?((?P<user>[^/]+)/)?(?P<model>[^.]*)(\.(?P<application>[^:]*(:.*)?))?)?`)
```
achilleasa: If you want you can make this stricter by concatenating the  "snippet" regexes from names.v2 (e.g https://github.com/juju/names/blob/v2/application.go#L13). I have used a similar approach for validating the endpoints: https://github.com/juju/charm/blob/v6/bundledata.go#L522-L527

offerurl.go
```diff
@@ -0,0 +1,170 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/juju/errors"
+	names "gopkg.in/juju/names.v2"
+)
+
+// OfferURL represents the location of an offered application and its
+// associated exported endpoints.
+type OfferURL struct {
```
achilleasa: Does this mean that we can strip off some code from juju/juju and use this parser instead?

bundledata.go
mitechie: Let's punt it for the moment. There's use cases where this will be something reused "relate me to an offer called 'keystone' on this controller" that we don't want to prevent.

offerurl.go
SimonRichardson: @achilleasa exactly

SimonRichardson:
$$merge$$

PR #282, opened 2019-06-26
Allow SAAS to work with relations

SimonRichardson:
The verifier needs to ensure that if the application name doesn't
exist in the application block, then check it in the SAAS block
also. That way we can still verify the bundle data ahead of time.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -756,7 +756,12 @@ func (verifier *bundleDataVerifier) verifyRelations() {
 				relParseErr = true
 				continue
 			}
-			if _, ok := verifier.bd.Applications[ep.application]; !ok {
+			// with the introduction of the SAAS block to bundles, we should
+			// test that not only is the expected application is in the
+			// applications block, but if it's not, is it in the SAAS offering.
+			_, foundApp := verifier.bd.Applications[ep.application]
+			_, foundSaas := verifier.bd.Saas[ep.application]
+			if !foundApp && !foundSaas {
 				verifier.addErrorf("relation %q refers to application %q not defined in this bundle", relPair, ep.application)
 			}
 			epPair[i] = ep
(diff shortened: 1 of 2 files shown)


SimonRichardson:
https://trello.com/c/0HjFzIMN/1583-cmr-allow-bundle-verification-when-using-saas-and-relations

bundledata_test.go
```diff
@@ -316,6 +316,36 @@ applications:
...
+		Applications: map[string]*charm.ApplicationSpec{
+			"wordpress": {
+				Charm:    "cs:trusty/wordpress-5",
+				NumUnits: 1,
+			},
+		},
+		Relations: [][]string{
+			{"wordpress:db", "mysql:db"},
+		},
+	},
```
achilleasa: For completeness, can you also add a verifier test for missing `saas` references?

bundledata.go
```diff
@@ -756,7 +756,12 @@ func (verifier *bundleDataVerifier) verifyRelations() {
 				relParseErr = true
 				continue
 			}
-			if _, ok := verifier.bd.Applications[ep.application]; !ok {
+			// with the introduction of the SAAS block to bundles, we should
+			// test that not only is the expected application is in the
+			// applications block, but if it's not, is it in the SAAS offering.
+			_, foundApp := verifier.bd.Applications[ep.application]
+			_, foundSaas := verifier.bd.Saas[ep.application]
+			if !foundApp && !foundSaas {
```
achilleasa: I think it would be nice to have an extra test `if foundApp && foundSaas` because in this case references would be ambiguous.

bundledata_test.go
SimonRichardson: it's not required :D

SimonRichardson:
$$merge$$

PR #283, opened 2019-06-27
SAAS and Application name clashes

SimonRichardson:
Ensure that we don't get application and SAAS name clashes when
verifying the bundle data


SimonRichardson:
$$merge$$

PR #284, opened 2019-07-03
Overlay merge rules

achilleasa:
This PR uses reflection to _re-implement_ the overlay -> base bundle merge logic that currently lives in the juju CLI. In my opinion, since the charms package contains the various bundle-related struct definitions it should also encapsulate the logic for applying overlays. In my view, the proposed implementation is a bit cleaner (less if blocks) and in contrast to the existing implementation it supports automatic merging of **all** application spec fields (e.g. trust, offers etc.).

## Working with multi-doc and standalone overlay files

The first part of the PR introduces a new set of abstractions for dealing with overlays which can now be either specified both as:
- **multi-doc** yaml documents (the new way)
- **standalone** yaml documents (the current way)

The key abstraction is a new interface called `BundleDataSource` which provides a `Parts` method that yields a list of `BundleDataPart` instances. Each `BbundleDataPart`  encapsulates a parsed `BundleData` instance and a field presence map which the overlay merging logic queries to figure out whether empty/nil `BundleData` fields were actually specified as such or were simply missing from the bundle data stream and therefore should be ignored when merging.

To make integration with the CLI code easier, the charms package provides two concrete implementations of the `BundleDataSource` interface:

- **LocalBundleDataSource** which reads a (potentially multi-doc) yaml file/folder/archive from disk.
- **StreamBundleDataSource** which reads a (potentially multi-doc) yaml stream from an `io.Reader`. This is primarily used for testing but external packages could leverage this for their own `BundleDataSource` implementations (e.g. to read data off an http response).

## Ok, so how do I get the final merged bundle?

The charms package provides a convenience helper function called `ReadAndMergeBundleData` that 
receives as input a slice of `BundleDataSource` instances. The function will fetch their contents, treat the **first** `bundleDataPart` as the base bundle and all other parts as overlays. The overlays will be sequentially applied to the base bundle and the final, merged bundle is returned back to the caller. 

Before the final bundle is returned to the caller, the helper will iterate the values of application options/annotations and machine annotations looking for the presence of special **include-directives** (identified via the prefix `include-file://` or `include-base64://`). Any include directives are resolved relative to the **base path** of the base bundle which is only non-empty for local bundle paths (yaml files or exploded archives) but _not_ for charm archives.

## Overlay merge rules

While the original CLI code seems to follow some straightforward patterns when merging maps, there are a few exceptions which need to be handled in a slightly different manner. This implementation attempts to re-produce the exact merge behavior and comes with a very verbose test to make sure everything works as expected ™️ 

When merging an overlay into a base bundle the following rules apply for the
BundleData struct fields:
- if an overlay specifies a bundle-level series, it overrides the base bundle
  series.
- overlay-defined relations are appended to the base bundle relations
- overlay-defined machines overwrite the base bundle machines.
- if an overlay defines an application that is not present in the base bundle,
  it will get appended to the application list.
- if an overlay defines an empty application value, it will be removed from base bundle together with any associated relations. For example, to remove an application named "mysql" the following overlay snippet can be  provided:
     ```yaml 
     applications:
        mysql:
     ```
- if an overlay defines an application that is also present in the base bundle
  the two application specs are merged together (see following rules)
- the above application-specific rules **also** apply when merging **sass** blocks.

ApplicationSpec merge rules:
- if the overlay defines a value for a scalar or slice field, it will overwrite
  the value from the base spec (e.g. trust, series etc).
- if the overlay specifies a nil/empty value for a map field, then the map
  field of the base spec will be cleared.
- if the overlay specifies a non-empty value for a map field, its key/value
  tuples are iterated and:
  - if the value is nil/zero and the type of the value is non-scalar, it is deleted from the base spec
  - otherwise, the key/value is inserted into the base spec overwriting any
    existing entries.

### DIFF
--- a/bundledatasrc.go
+++ b/bundledatasrc.go
@@ -1,0 +1,186 @@
+// Copyright 2019 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package charm
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/juju/errors"
+	"gopkg.in/yaml.v2"
+)
+
+// FieldPresenseMap indicates which keys of a parsed bundle yaml document were
+// present when the document was parsed. This map is used by the overlay merge
+// code to figure out whether empty/nil field values were actually specified as
+// such in the yaml document.
+type FieldPresenseMap map[interface{}]interface{}
+
+func (fpm FieldPresenseMap) fieldPresent(fieldName string) bool {
+	_, exists := fpm[fieldName]
+	return exists
+}
+
+func (fpm FieldPresenseMap) forField(fieldName string) FieldPresenseMap {
+	v, exists := fpm[fieldName]
+	if !exists {
+		return nil
+	}
+
+	asMap, valid := v.(FieldPresenseMap)
+	if !valid {
+		panic(errors.Errorf("field map entry %q does not point to a nested field presense map", fieldName))
+	}
+	return asMap
+}
+
+// BundleDataPart combines a parsed BundleData instance with a nested map that
+// can be used to discriminate between fields that are missing from the data
+// and those that are present but defined to be empty.
+type BundleDataPart struct {
+	Data        *BundleData
+	PresenseMap FieldPresenseMap
+}
+
+// BundleDataSource is implemented by types that can parse bundle data into a
+// list of composable parts.
+type BundleDataSource interface {
+	Parts() []*BundleDataPart
+}
+
+type resolvedBundleDataSource struct {
+	basePath string
+	parts    []*BundleDataPart
+}
+
+func (s *resolvedBundleDataSource) Parts() []*BundleDataPart {
+	return s.parts
+}
+
+// LocalBundleDataSource reads a (potentially multi-part) bundle from path and
+// returns a BundleDataSource for it. Path may point to a yaml file, a bundle
+// directory or a bundle archive.
+func LocalBundleDataSource(path string) (BundleDataSource, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		if isNotExistsError(err) {
+			return nil, errors.NotFoundf("%q", path)
+		}
+
+		return nil, errors.Annotatef(err, "stat failed for %q", path)
+	}
+
+	// Treat as an exploded bundle archive directory
+	if info.IsDir() {
+		path = filepath.Join(path, "bundle.yaml")
+	}
+
+	// Try parsing as a yaml file first
+	f, err := os.Open(path)
+	if err != nil {
+		if isNotExistsError(err) {
+			return nil, errors.NotFoundf("%q", path)
+		}
+		return nil, errors.Annotatef(err, "access bundle data at %q", path)
+	}
+	defer func() { _ = f.Close() }()
+
+	parts, pErr := parseBundleParts(f)
+	if pErr == nil {
+		absPath, err := filepath.Abs(path)
+		if err != nil {
+			return nil, errors.Annotatef(err, "resolve absolute path to %s", path)
+		}
+		return &resolvedBundleDataSource{
+			basePath: filepath.Dir(absPath),
+			parts:    parts,
+		}, nil
+	}
+
+	// As a fallback, try to parse as a bundle archive
+	zo := newZipOpenerFromPath(path)
+	zrc, err := zo.openZip()
+	if err != nil {
+		// Not a zip file; return the original parse error
+		return nil, errors.NewNotValid(pErr, "cannot unmarshal bundle contents")
+	}
+	defer func() { _ = zrc.Close() }()
+
+	r, err := zipOpenFile(zrc, "bundle.yaml")
+	if err != nil {
+		// It is a zip file but not one that contains a bundle.yaml
+		return nil, errors.NotFoundf("interpret bundle contents as a bundle archive: %v", err)
+	}
+	defer func() { _ = r.Close() }()
+
+	if parts, pErr = parseBundleParts(r); pErr == nil {
+		return &resolvedBundleDataSource{
+			basePath: "", // use empty base path for archives
+			parts:    parts,
+		}, nil
+	}
+
+	return nil, errors.NewNotValid(pErr, "cannot unmarshal bundle contents")
+}
+
+func isNotExistsError(err error) bool {
+	if os.IsNotExist(err) {
+		return true
+	}
+	// On Windows, we get a path error due to a GetFileAttributesEx syscall.
+	// To avoid being too proscriptive, we'll simply check for the error
+	// type and not any content.
+	if _, ok := err.(*os.PathError); ok {
+		return true
+	}
+	return false
+}
+
+// StreamBundleDataSource reads a (potentially multi-part) bundle from r and
+// returns a BundleDataSource for it.
+func StreamBundleDataSource(r io.Reader, basePath string) (BundleDataSource, error) {
+	parts, err := parseBundleParts(r)
+	if err != nil {
+		return nil, errors.NotValidf("cannot unmarshal bundle contents: %v", err)
+	}
+
+	return &resolvedBundleDataSource{parts: parts, basePath: basePath}, nil
+}
+
+func parseBundleParts(r io.Reader) ([]*BundleDataPart, error) {
+	b, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		// Ideally, we would be using a single reader and we would
+		// rewind it to read each block in structured and raw mode.
+		// Unfortunately, the yaml parser seems to parse all documents
+		// at once so we need to use two decoders.
+		structDec = yaml.NewDecoder(bytes.NewReader(b))
+		rawDec    = yaml.NewDecoder(bytes.NewReader(b))
+		parts     []*BundleDataPart
+	)
+
+	for docIdx := 0; ; docIdx++ {
+		var part BundleDataPart
+
+		err = structDec.Decode(&part.Data)
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, errors.Annotatef(err, "unmarshal document %d", docIdx)
+		}
+
+		// We have already checked for errors for the previous unmarshal attempt
+		_ = rawDec.Decode(&part.PresenseMap)
+		parts = append(parts, &part)
+	}
+
+	return parts, nil
+}
--- a/bundledatasrc.go
+++ b/bundledatasrc.go
@@ -50,6 +50,8 @@ type BundleDataPart struct {
 // list of composable parts.
 type BundleDataSource interface {
 	Parts() []*BundleDataPart
+	BasePath() string
+	ResolveInclude(path string) ([]byte, error)
 }
 
 type resolvedBundleDataSource struct {
@@ -59,6 +61,41 @@ type resolvedBundleDataSource struct {
 
 func (s *resolvedBundleDataSource) Parts() []*BundleDataPart {
 	return s.parts
+}
+
+func (s *resolvedBundleDataSource) BasePath() string {
+	return s.basePath
+}
+
+func (s *resolvedBundleDataSource) ResolveInclude(path string) ([]byte, error) {
+	absPath := path
+	if !filepath.IsAbs(absPath) {
+		var err error
+		absPath, err = filepath.Abs(filepath.Clean(filepath.Join(s.basePath, absPath)))
+		if err != nil {
+			return nil, errors.Annotatef(err, "resolving relative include %q", path)
+		}
+	}
+
+	info, err := os.Stat(absPath)
+	if err != nil {
+		if isNotExistsError(err) {
+			return nil, errors.NotFoundf("include file %q", path)
+		}
+
+		return nil, errors.Annotatef(err, "stat failed for %q", path)
+	}
+
+	if info.IsDir() {
+		return nil, errors.Errorf("include path %q resolves to a folder", path)
+	}
+
+	data, err := ioutil.ReadFile(absPath)
+	if err != nil {
+		return nil, errors.Annotatef(err, "reading include file at %q", path)
+	}
+
+	return data, nil
 }
 
 // LocalBundleDataSource reads a (potentially multi-part) bundle from path and
--- a/bundledatasrc_test.go
+++ b/bundledatasrc_test.go
@@ -5,6 +5,7 @@ package charm
 
 import (
 	"archive/zip"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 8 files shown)


bundledatasrc.go
```diff
@@ -0,0 +1,109 @@
...
+	"bytes"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/juju/errors"
+	"gopkg.in/yaml.v2"
+)
+
+type fieldPresenseMap map[interface{}]interface{}
```
jameinel: if  fieldName is always a string, why isn't this a map[string]interface{} ?

bundledatasrc.go
```diff
@@ -0,0 +1,109 @@
...
+// and those that are present but defined to be empty.
+type bundleDataPart struct {
+	data        *BundleData
+	presenseMap fieldPresenseMap
+}
+
+// BundleDataSource is implemented by types that can parse bundle data into a
+// list of composable parts.
+type BundleDataSource interface {
+	read() ([]*bundleDataPart, error)
```
jameinel: Do you want to make read a private method?

bundledatasrc.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// FileBundleDataSource is a BundleDataSource that reads bundle data from a file.
+type FileBundleDataSource string
+
+// read implements BundleDataSource.
+func (s FileBundleDataSource) read() ([]*bundleDataPart, error) {
```
jameinel: Arguably this is a filename bundledatasource because if you had a File that fits as an io.Reader and you would just use StreamBundleDataSource, right?
It seems a bit odd to me that FileBundleDataSource *is a* string while StreamBundleDataSource *has a* reader. I'd rather they were consistent about whether they are actually a struct or not.

bundledatasrc.go
```diff
@@ -0,0 +1,109 @@
...
+
+		err = structDec.Decode(&part.data)
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, errors.Annotatef(err, "unmarshal document %d from %v", docIdx, s)
+		}
+
+		// We have already checked for errors for the previous unmarshal attempt
+		_ = rawDec.Decode(&part.presenseMap)
```
jameinel: This feels a lot like it should actually be a change to the YAML Decode contract that would allow us to decode into a struct, but also ask the decoder what attributes were actually there. We can live with this, but what do you think about trying to update Decode to something like DecodeWithKeys

bundledatasrc_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (s *BundleDataSourceSuite) TestReadMultiDocBundle(c *gc.C) {
+	path := bundleDirPath(c, "wordpress-multidoc")
+	ds := FileBundleDataSource(filepath.Join(path, "bundle.yaml"))
+
+	parts, err := ds.read()
+	c.Assert(err, gc.IsNil)
+	c.Assert(parts, gc.HasLen, 3)
+	s.assertFieldPresent(c, parts[1], "applications.wordpress.offers.offer1.endpoints")
+	s.assertFieldPresent(c, parts[2], "applications.wordpress.offers.offer1.acl.admin")
```
jameinel: don't you want to both assert the positive, and the negative? (assertFieldNotPresent) ?

bundledatasrc_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+type BundleDataSourceSuite struct {
+	testing.IsolationSuite
+}
+
+var _ = gc.Suite(&BundleDataSourceSuite{})
+
+func (s *BundleDataSourceSuite) TestReadMultiDocBundle(c *gc.C) {
+	path := bundleDirPath(c, "wordpress-multidoc")
+	ds := FileBundleDataSource(filepath.Join(path, "bundle.yaml"))
```
jameinel: the decoupling is ok, but it does make it quite a bit harder to figure out what this test is actually doing. The bundle.yaml doesn't look that big, is it reasonable to change this into:

const MultiDocBundle = `
...
`[1:]

...
  ds := StreamBundleDataSource(bytes.Buffer(MultiDocBundle))

?

overlay.go
```diff
@@ -331,3 +335,321 @@ func isZero(v reflect.Value) bool {
...
+// to resolve any include directives present in the machine annotations,
+// application options and annotations by invoking the provided includeResolver
+// with the path to the file required for each include.
+//
+// When merging an overlay into a base bundle the following rules apply for the
+// BundleData struct fields:
+// - if an overlay specifies a bundle-level series, it overrides the base bundle
+//   series.
+// - overlay-defined relations are appended to the base bundle relations
+// - overlay-defined machines overwrite the baase bundle machines.
```
jameinel: s/baase/base/

overlay.go
```diff
@@ -331,3 +335,321 @@ func isZero(v reflect.Value) bool {
...
+// - if an overlay defines an application that is not present in the base bundle,
+//   it will get appended to the application list.
+// - if an overlay defines an empty/nil application, it will be removed from the
```
jameinel: if an overlay defines an application with no contents ?
Otherwise it sounds like the app is nil, not the contents of the app declaration. I'm not sure if it is a lot better.

overlay.go
```diff
@@ -331,3 +335,321 @@ func isZero(v reflect.Value) bool {
...
+		if err := applyOverlay(base, overlay); err != nil {
+			return nil, err
+		}
+	}
+
+	if includeResolver == nil {
+		return base.data, nil
+	}
+
+	for app, appData := range base.data.Applications {
```
jameinel: is it ok to resolve includes after applying overlays? I would hope thats ok, just checking as I think it is a change in logic.
(previously I think you couldn't put includes in overlays?)

bundledatasrc.go
achilleasa: That's a good observation. I will change it to use string keys instead.

bundledatasrc.go
achilleasa: The initial thought was to only allow BundleDataSource implementations inside the charms package (this is also why bundleDataPart is un-exported). Given that we probably need to add a BundleDataSource implementation for dealing with remote bundles into another bundle I will make these public.

bundledatasrc.go
achilleasa: Also a good observation. As I will be moving the "is-this-a-bundle-or-a-dir-or-an-archive" logic into this type I will also be renaming it to something like a "LocalPathBundleDataSource" and convert it into a struct as we also need to track the base-path for resolving includes.

bundledatasrc.go
achilleasa: Interesting idea. I also thought it would be great not to have to do the decode dance twice. I believe we can land the current set of changes, then open a PR for the yaml package and address this shortcoming in the future. What do you think?

bundledatasrc_test.go
achilleasa: Sure thing. This particular test exercises the FileBundleDataSource implementation. I will update this together with my changes for making the FileBundleDataSource work with local paths and/or archives (see comment above)

overlay.go
achilleasa: The proposed implementation can work with both ways because the removal predicate is:
`defined-in-overlay-field-map && isZeroOrNil(val)`.

overlay.go
achilleasa: This behavior matches the current implementation: https://github.com/juju/juju/blob/develop/cmd/juju/application/bundle.go#L76

However, the current implementation _seems to assume_ that only the base bundle may contain include directives as the code resolves include paths relative to the base bundle path (https://github.com/juju/juju/blob/develop/cmd/juju/application/bundle.go#L143) which kinda limits the usefulness of having includes in overlays.

bundledatasrc.go
achilleasa: Actually, now that I thought about it again, I am using this because it's the default map type used by the yaml package (https://github.com/go-yaml/yaml/blob/v2/decode.go#L237). If I change this to use string keys, nested maps will use interface{} keys.

bundledatasrc.go
jameinel: works for me

overlay.go
jameinel: My point is more just phrasing on the doc. "what is a nil application". Maybe others wouldn't find the wording confusing.

bundledatasrc_test.go
achilleasa: Actually, there isn't any value in testing the negative as the yaml parser will simply not populate the map if the keys do not exist.

bundledatasrc_test.go
```diff
@@ -0,0 +1,214 @@
...
+    offers:
+      offer1:
+        acl:
+          admin: "admin"
+          foo: "consume"
+`
+)
+
+func (s *BundleDataSourceSuite) TestReadMultiDocBundle(c *gc.C) {
+	specs := []struct {
```
jameinel: In my experience, table based tests like this aren't worth it. They make it too hard to just run one of the subtests, and they make it harder to look at the control flow. Better to have:
TestReadMultiDocBundleYaml(c *gc.C) {
  path := bundleDirPath(c, "wordpress-multidoc")
  source := LocalBundleDataSource(...)
  assertBundleSourceProcessed(c, source)
}
And then repeat that for
TestReadMultiDocBundleArchive
TestReadMultiDocBundleDirectory
TestReadMultiDocBundleStream

You get to still share all the same code via a helper function, and it makes the tests much easier to manage.

bundledatasrc_test.go
```diff
@@ -0,0 +1,214 @@
...
+		s.assertFieldPresent(c, parts[1], "applications.wordpress.offers.offer1.endpoints")
+		s.assertFieldPresent(c, parts[2], "applications.wordpress.offers.offer1.acl.admin")
+	}
+}
+
+func (s *BundleDataSourceSuite) TestResolveAbsoluteFileInclude(c *gc.C) {
+	target, err := filepath.Abs(filepath.Join(c.MkDir(), "example"))
+	c.Assert(err, gc.IsNil)
+
+	expContent := []byte("example content")
```
jameinel: small aside, but "\n" at the end makes content a lot easier to work with, I think.

bundledatasrc_test.go
```diff
@@ -0,0 +1,214 @@
...
+	c.Assert(ioutil.WriteFile(target, expContent, os.ModePerm), gc.IsNil)
+
+	ds := new(resolvedBundleDataSource)
+
+	got, err := ds.ResolveInclude(target)
+	c.Assert(err, gc.IsNil)
+	c.Assert(got, gc.DeepEquals, expContent)
```
jameinel: we probably also would rather compare (string(got)) because otherwise it tells you [0x32, 0x23,} != [0x12, 0x45]

which is usually not very nice to understand as a human. :)

bundledatasrc_test.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (s *BundleDataSourceSuite) TestResolveRelativeFileInclude(c *gc.C) {
+	relTo := c.MkDir()
+	target, err := filepath.Abs(filepath.Join(relTo, "example"))
+	c.Assert(err, gc.IsNil)
+
+	expContent := []byte("example content")
```
jameinel: ditto

bundledatasrc_test.go
```diff
@@ -0,0 +1,214 @@
...
+			incPath: tmpDir,
+			exp:     fmt.Sprintf("include path %q resolves to a folder", tmpDir),
+		},
+	}
+
+	ds := new(resolvedBundleDataSource)
+	for specIndex, spec := range specs {
+		c.Logf("[test %d] %s", specIndex, spec.descr)
+
+		_, err := ds.ResolveInclude(spec.incPath)
```
jameinel: do you feel this is easier to maintain/add more than just separate tests? This is borderline for me, so maybe it isn't worth changing anything. But you do end up with log messages that are too long, and extra messages so you have a chance to figure out what is going on. But the line you need to edit *isn't* the line that c.Assert is on. (you have to figure out which test spec needs updating)

overlay.go
```diff
@@ -331,3 +338,331 @@ func isZero(v reflect.Value) bool {
...
+		if base.Data.Saas == nil {
+			base.Data.Saas = make(map[string]*SaasSpec, len(overlay.Data.Saas))
+		}
+
+		fpm := overlay.PresenceMap.forField("saas")
+		for srcSaasName, srcSaasSpec := range overlay.Data.Saas {
+			// If the overlay map points to an empty object, delete
+			// it from the base bundle
+			if isZero(reflect.ValueOf(srcSaasSpec)) {
+				delete(base.Data.Saas, srcSaasName)
```
jameinel: sholudn't we include the SAAS entries for "removeRelations" ? (if you remove a SAAS entry, then you should remove the relations to that SAAS entry, just like application)

overlay.go
```diff
@@ -331,3 +338,331 @@ func isZero(v reflect.Value) bool {
...
+
+			// if this is a new saas block just append it; otherwise
+			// recursively merge the two saas specs.
+			dstSaasSpec, defined := base.Data.Saas[srcSaasName]
+			if !defined {
+				base.Data.Saas[srcSaasName] = srcSaasSpec
+				continue
+			}
+
+			mergeStructs(dstSaasSpec, srcSaasSpec, fpm.forField(srcSaasName))
```
jameinel: We should probably be including a step that means if you define both an Application named 'foo' and a SAAS named 'foo' then the bundle is incorrect, since I'm pretty sure you can't have both in a model.

overlay.go
```diff
@@ -331,3 +338,331 @@ func isZero(v reflect.Value) bool {
...
+			}
+
+			dstMap := dst.Field(i)
+			if dstMap.IsNil() {
+				dstMap.Set(reflect.MakeMap(srcVal.Type()))
+			}
+			for _, srcKey := range srcVal.MapKeys() {
+				// If the key points to an empty value delete it from the dst map
+				srcMapVal := srcVal.MapIndex(srcKey)
+				if isZero(srcMapVal) {
```
jameinel: Does this mean that:
{
 "foo": 0
}

will delete foo?
I'm wondering about config items where you might want to actually set a value to 0. Thoughts?
I'm just wondering about cases where there is a default value, so setting the value to 0 or "" is different than not setting it at all.

overlay_test.go
```diff
@@ -181,3 +185,431 @@ series: bionic
 	sort.Strings(expErrors)
 	c.Assert(errStrings, jc.DeepEquals, expErrors)
 }
+
+func (*bundleDataOverlaySuite) TestReadAndMergeBundleData(c *gc.C) {
+	data := `
```
jameinel: Would this be easier to understand if we broke it into several tests? ReadAndMergeBundleDataChangeCharm
ReadAndMergeBundleDataChangeSeries, etc.?
it takes a lot for me to read through this test and really map all the things that the comments say into things that it is actually doing, and track that it isn't doing anything that *isn't* commented.

Thoughts?

overlay.go
achilleasa: This case is handled when the bundle is verified: https://github.com/juju/charm/blob/v6/bundledata.go#L542

overlay.go
achilleasa: That is correct and I hadn't really thought about this case. However we _do_ need this functionality for _some_ of the maps (`offers` within `ApplicationSpec` being one example). 

I wonder if we could address this problem by checking whether the value is a scalar and only applying the delete logic for object values. What do you think?

overlay_test.go
achilleasa: Yeah... I admit this one is a horribly verbose test but I wanted to have a test that checked all possible the mutations in a single block as it could help catch subtle bugs that don't manifest when checking each mutation in isolation.

Having said that, I don't mind to make the changes you propose especially if it makes reading the tests easier.

overlay.go
achilleasa: Good catch!

bundledatasrc_test.go
achilleasa: While personally I prefer table-driven tests (wrapped in t.Run when using just the native go test stuff) I do understand that it's cumbersome to work with them using gocheck. Generally, when I am adding code I try to emulate the code/test style that is already present in the files I tough and IIRC charms has a few table-driven tests.

bundledatasrc_test.go
```diff
@@ -46,77 +67,40 @@ applications:
...
+	c.Assert(string(got), gc.DeepEquals, string(expContent))
 }
 
 func (s *BundleDataSourceSuite) TestResolveRelativeFileInclude(c *gc.C) {
 	relTo := c.MkDir()
 	target, err := filepath.Abs(filepath.Join(relTo, "example"))
 	c.Assert(err, gc.IsNil)
 
-	expContent := []byte("example content")
+	expContent := []byte("example content\n")
```
jameinel: doesn't need to be changed, but I would usually have the 'canonical' form as a string, and then convert to []byte only where you need it. eg:

expContent := "example content\n"
cAssert(ioutil.WriteFile(target, []byte(expContent), os.ModePerm), gc.IsNil)

...
c.Assert(string(got), gc.Equals, expContent)

(don't need DeepEquals for strings)

achilleasa: Wow, how did I miss that? Fixed now!

achilleasa:
$$merge$$

PR #286, opened 2019-07-25
Delete keys for non-retained entries when splitting bundles

achilleasa:
The problem manifests when trying to split a bundle that contains two applications where **only one** exposes a set of offers:

```yaml
applications:
  mysql:
    charm: cs:mysql
    offers:
      my-offer:
        endpoints:
        - apache-website
        - website-cache
  wordpress:
    charm: cs:wordpress
```

The visitors used in the split bundle to base/overlay logic would clear all non-overly fields in the `mysql` and `wordpress` and report that we need to retain the `mysql` block (contains overlay bits) and not the `wordpress` block. Because **at least** one of the `application` map keys must be retained, the `application` map and its ancestors must also be retained.

As a result if we encode the overlay we would get this:

```yaml
applications:
  mysql:
    offers:
      my-offer:
        endpoints:
        - apache-website
        - website-cache
  wordpress: {}
```
If we were to merge that overlay, the merge code would erroneously assume that the wordpress app has to be deleted!

The fix for this involves adding a flag to the context shared between the various visitors that specifies the policy for retaining non-required map keys. The bundle split code opts to drop non-required keys if the map value is a non-scalar type (i.e you can drop application entries but not options map values that are set to 0).

### DIFF
--- a/overlay.go
+++ b/overlay.go
@@ -70,10 +70,16 @@ import (
 // data and can thus be safely manipulated by the caller.
 func ExtractBaseAndOverlayParts(bd *BundleData) (base, overlay *BundleData, err error) {
 	base = cloneBundleData(bd)
-	_ = visitField(&visitorContext{structVisitor: clearOverlayFields}, base)
+	_ = visitField(&visitorContext{
+		structVisitor:          clearOverlayFields,
+		dropNonRequiredMapKeys: false,
+	}, base)
 
 	overlay = cloneBundleData(bd)
-	_ = visitField(&visitorContext{structVisitor: clearNonOverlayFields}, overlay)
+	_ = visitField(&visitorContext{
+		structVisitor:          clearNonOverlayFields,
+		dropNonRequiredMapKeys: true,
+	}, overlay)
 
 	return base, overlay, nil
 }
@@ -155,6 +161,8 @@ type visitorContext struct {
 	// An optional pre/post visitor for indexable items (slices, maps)
 	indexedElemPreVisitor  func(index interface{})
 	indexedElemPostVisitor func(index interface{})
+
+	dropNonRequiredMapKeys bool
 }
 
 // visitField invokes ctx.structVisitor(val) if v is a struct and returns back
@@ -200,7 +208,19 @@ func visitFieldsInMap(ctx *visitorContext, val reflect.Value) (result bool) {
 			ctx.indexedElemPreVisitor(key)
 		}
 
-		result = visitField(ctx, v.Interface()) || result
+		visRes := visitField(ctx, v.Interface())
+		result = visRes || result
+
+		// If the map value is a non-scalar value and the visitor
+		// returned false (don't retain), consult the dropNonRequiredMapKeys
+		// hint to decide whether we need to delete the key from the map.
+		//
+		// This is required when splitting bundles into base/overlay
+		// bits as empty map values would be encoded as empty objects
+		// that the overlay merge code would mis-interpret as deletions.
+		if !visRes && isNonScalar(v) && ctx.dropNonRequiredMapKeys {
+			val.SetMapIndex(key, reflect.Value{})
+		}
 
 		if ctx.indexedElemPostVisitor != nil {
 			ctx.indexedElemPostVisitor(key)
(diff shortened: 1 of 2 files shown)


achilleasa:
!!build!!

(DNS issues)

overlay.go
```diff
@@ -200,7 +208,19 @@ func visitFieldsInMap(ctx *visitorContext, val reflect.Value) (result bool) {
...
+		result = visRes || result
+
+		// If the map value is a non-scalar value and the visitor
+		// returned false (don't retain), consult the dropNonRequiredMapKeys
+		// hint to decide whether we need to delete the key from the map.
+		//
+		// This is required when splitting bundles into base/overlay
+		// bits as empty map values would be encoded as empty objects
+		// that the overlay merge code would mis-interpret as deletions.
+		if !visRes && isNonScalar(v) && ctx.dropNonRequiredMapKeys {
```
jameinel: I don't quite understand why this works. It feels like it should be !ctx.dropNonRequired...
because this is "I don't have a value, but the type isn't scalar". eg I have a map.
Which means we shouldn't set *anything* if we are wanting to drop non-required keys. Otherwise we will mark the keys for deletion.

Maybe I'm misinterpreting the definition of "dropNonRequiredMapKeys" ?

overlay_test.go
```diff
@@ -138,6 +138,92 @@ relations:
...
+
+	bd, err := charm.ReadBundleData(strings.NewReader(data))
+	c.Assert(err, gc.IsNil)
+
+	base, overlay, err := charm.ExtractBaseAndOverlayParts(bd)
+	c.Assert(err, jc.ErrorIsNil)
+
+	baseYaml, err := yaml.Marshal(base)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert("\n"+string(baseYaml), gc.Equals, expBase)
```
jameinel: I'd rather express this as:
```
expected := `
blah
blah
`[1:]
```
rather than "\n" + got.
It is a common pattern we use elsewhere, so I think people will understand and expect it.

overlay.go
achilleasa: The `visitField` implementation checks the type of `v` and invokes specialized methods for working with maps, slices or structs. Otherwise, it returns `false` which is interpreted as "I don't really know how to decide if you need scalar values so I just return false to be on the safe-side". This assumes that the parent container (map, slice, pointer to some object in the struct) is properly annotated so we can retain it if we need to (the offers map for instance)

An easy example to visualize this is to think of the struct processing code operating on a `BundleData` instance and currently processing the `ApplicationMap`. So eventually, the `visitFieldsInMap` gets called. We iterate the map and recursively visit each `*Application`. entry in the map. The first visitor returns true (it has an offer) and the second returns false. 

The predicate here should be interpreted as: `I don't need to retain the map entry for this key (visitor returned false) AND the type of the value is non-scalar AND I was asked to drop these kinds of keys from the map`. 

Essentially, the implementation assumes that the value which should probably not be retained points to an object. This means the struct processing code must have executed and we can safely trust the result given that only struct processing code checks for overlay-related annotations.

Hope my explanation helps to answer your question.

overlay_test.go
achilleasa: Sure thing.

overlay.go
```diff
@@ -70,10 +70,16 @@ import (
 // data and can thus be safely manipulated by the caller.
 func ExtractBaseAndOverlayParts(bd *BundleData) (base, overlay *BundleData, err error) {
 	base = cloneBundleData(bd)
-	_ = visitField(&visitorContext{structVisitor: clearOverlayFields}, base)
+	_ = visitField(&visitorContext{
+		structVisitor:          clearOverlayFields,
+		dropNonRequiredMapKeys: false,
```
achilleasa: By virtue of copy-paste I had initially set this to true which is incorrect. It can actually clear the contents of maps such as `options` so it is only required for the overlay bit.

I have now fixed this and added an option block to the relevant test to verify that it works as expected for the base bundle.

overlay.go
jameinel: It was just that I didn't really Set(..., reflect.Value{}) is actually Delete(). LGTM

achilleasa:
$$merge$$

PR #287, opened 2019-07-25
Add method to bundle for reporting presence of overlays

achilleasa:
This PR extends the Bundle interface with a new method `ContainsOverlays` that users of the charm package can use to check whether the bundle they just obtained using one of the Read helpers did in fact contain one or more overlay blocks that _were not actually parsed_ as the Read helpers do not work with multi-document bundles.

This new functionality will be leveraged by the charmstore code to reject uploads for multi-doc bundles.


bundle.go
```diff
@@ -15,8 +15,10 @@ import (
 type Bundle interface {
 	// Data returns the contents of the bundle's bundle.yaml file.
 	Data() *BundleData
-	// Data returns the contents of the bundle's README.md file.
+	// ReadMe returns the contents of the bundle's README.md file.
```
jameinel: 👍

bundlearchive_test.go
```diff
@@ -36,6 +36,18 @@ func (s *BundleArchiveSuite) TestReadBundleArchiveBytes(c *gc.C) {
...
+}
+
+func (s *BundleArchiveSuite) TestReadMultiDocBundleArchiveBytes(c *gc.C) {
+	path := archivePath(c, readBundleDir(c, "wordpress-simple-multidoc"))
+	data, err := ioutil.ReadFile(path)
+	c.Assert(err, gc.IsNil)
+
+	archive, err := charm.ReadBundleArchiveBytes(data)
+	c.Assert(err, gc.IsNil)
+	c.Assert(archive.ContainsOverlays(), gc.Equals, true)
```
jameinel: There is "jc.IsTrue" and "jc.IsFalse" which gives slightly better output when it doesn't match (though maybe we don't depend on github.com/juju/testing/checkers here)

achilleasa: I guess it's just muscle memory :D 

bundle_test.go does import jc so since we already depend on checkers I will update my code to use that instead.

achilleasa:
@jameinel Yes, I will be following up with a charmstore PR that uses the new helper method.

bundlearchive_test.go
achilleasa: Pushed https://github.com/juju/charm/pull/287/commits/1d63110874911321b8ac69154e247fee7c728d6f to address this comment.

achilleasa:
$$merge$$

PR #289, opened 2019-08-06
Resolve relative charm paths while merging bundle overlays

achilleasa:
This PR adds a little bit of extra logic to the bundle merge code so we can properly resolve local charm paths relative to the bundle or overlay that defines them.


achilleasa:
$$mege$$

PR #290, opened 2019-08-27
Prevent panic in visitField when a nil value is passed as an argument

achilleasa:
Calling reflect.ValueOf() on a nil value and then invoking Kind() on the
returned reflected Type causes a panic.


achilleasa:
$$merge$$

PR #291, opened 2019-09-13
Add a few more fields to Deployment in metadata.yaml

ycliuhw:
Add omit type for Deployment.ServiceType, new fields Daemonset and MinVersion;


ycliuhw:
$$merge$$

PR #294, opened 2019-10-18
Refactor version handling for charmdir usage

nammn:
## Summary
bugfix: adds version to charmdir interface
charmdir: adds version file parsing to
refactor: version parsing is done from the beginning instead during archiveTo

## Disclaimer
one thing I am not 100% sure is, by extracting the version generation from "archiveTo" to the initial "charmdir" creation, any updates which happen between "creation" and "archiveTo" would not be added. I am not sure whether there can be any updates anyway and should be. I will add this to the PR as "disclaimer" to think about it while going through.

One could be 100% sure and just recreate the `versionstring` by calling the function again in `archiveTo`. For now I did not include this as part of discussion.

I did, on purpose, not add the Version method to the interface as this would mean to change multiple other repos as. It would make sense but it would lead to a lot of pr's which I wanted to keep out for now.

## Description
Currently, the version for `charmDir` only gets created when `archiveTo` is called.
There can be cases that we have no vcs, but a version file located like in the `charmArchives`

This PR should address this issue by doing:
- refactor when we create the version number 
- add a way to parse the version file
- add a more sophisticated error message

### DIFF
--- a/charmdir.go
+++ b/charmdir.go
@@ -49,6 +49,7 @@ type CharmDir struct {
 	actions    *Actions
 	lxdProfile *LXDProfile
 	revision   int
+	version    string
 }
 
 // Trick to ensure *CharmDir implements the Charm interface.
@@ -119,6 +120,12 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
 			return nil, errors.New("invalid revision file")
 		}
 	}
+	version, _, err := dir.MaybeGenerateVersionString(logger)
+	if err != nil {
+		// We don't want to stop, even if the version cannot be generated
+		logger.Errorf("unexpected problem trying to generate version string: %q", err)
+	}
+	dir.version = version
 
 	file, err = os.Open(dir.join("lxd-profile.yaml"))
 	if _, ok := err.(*os.PathError); ok {
@@ -177,6 +184,11 @@ func (dir *CharmDir) join(parts ...string) string {
 // expanded in dir.
 func (dir *CharmDir) Revision() int {
 	return dir.revision
+}
+
+// Version returns the VCS version representing the version file from archive.
+func (dir *CharmDir) Version() string {
+	return dir.version
 }
 
 // Meta returns the Meta representing the metadata.yaml file
@@ -246,21 +258,15 @@ func resolveSymlinkedRoot(rootPath string) (string, error) {
 // ArchiveTo creates a charm file from the charm expanded in dir.
 // By convention a charm archive should have a ".charm" suffix.
 func (dir *CharmDir) ArchiveTo(w io.Writer) error {
-	versionString, vcsType, err := dir.MaybeGenerateVersionString()
-	if err != nil {
-		// Just to be safe, ensure version is "" on error.
-		versionString = ""
-		logger.Warningf(
-			"%q version string generation failed : %v\nThis means that the charm version won't show in juju status.",
-			vcsType, err)
-	}
-
 	ignoreRules, err := dir.buildIgnoreRules()
 	if err != nil {
 		return err
 	}
+	var dummyLogger = NopLogger{}
+	// We update the version to make sure we don't lag behind
+	dir.version, _, _ = dir.MaybeGenerateVersionString(dummyLogger)
 
-	return writeArchive(w, dir.Path, dir.revision, versionString, dir.Meta().Hooks(), ignoreRules)
+	return writeArchive(w, dir.Path, dir.revision, dir.version, dir.Meta().Hooks(), ignoreRules)
 }
 
 func writeArchive(w io.Writer, path string, revision int, versionString string, hooks map[string]bool, ignoreRules ignoreRuleset) error {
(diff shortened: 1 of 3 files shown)


charmdir.go
```diff
@@ -119,6 +120,8 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
 			return nil, errors.New("invalid revision file")
 		}
 	}
+	version, _, err := dir.MaybeGenerateVersionString()
+	dir.version = version
```
SimonRichardson: So this is the meat of it, in that, we always read the version string (git or file) on reading the charm directory?

charmdir.go
```diff
@@ -425,9 +424,16 @@ func (dir *CharmDir) MaybeGenerateVersionString() (string, string, error) {
 		cmdArgs = []string{"bzr", "version-info"}
 	} else if _, err = os.Stat(filepath.Join(dir.Path, ".git")); err == nil {
 		cmdArgs = []string{"git", "describe", "--dirty", "--always"}
+	} else if file, err := os.Open(dir.join("version")); err == nil {
+		logger.Debugf("charm is not in version control and uses a version file")
+		var versionNumber = ""
+		_, err = fmt.Fscan(file, &versionNumber)
```
SimonRichardson: I wonder if we use `fmt.Fscanf` instead, so that we only grab the number and nothing more?

```sh
var versionNumber string
n, err = fmt.Fscan(file, "%s", &versionNumber)
if err != nil {
	return "", "", err
}
if n != 1 {
	return "", "", errors.Errorf("unexpected a version")
}
```

charmdir.go
```diff
@@ -438,17 +444,11 @@ func (dir *CharmDir) MaybeGenerateVersionString() (string, string, error) {
...
-	if err == nil {
-		return output, vcsType, nil
-	}
-
-	// If there's an error, we may be able to get a relevant cause string.
-	exitErr, ok := err.(*exec.ExitError)
-	if !ok || len(exitErr.Stderr) == 0 {
-		return "", vcsType, err
-	}
-	return "", vcsType, fmt.Errorf("%s: %s", err, string(exitErr.Stderr))
```
SimonRichardson: Good catch, this didn't do anything. I wonder why the deadcode didn't pick this up...?

charmdir.go
SimonRichardson: Also handling the `err` from `fmt.Fscan`

charmdir.go
nammn: @SimonRichardson yes one of it, right. I thought this would make sense, as we initially create the `CharmDir` struct. With this, we can make sure of consistent output. 
Additionally, I added this to the struct of the charmdir, because version is a field which should belong to the struct and not only on archiving IMO.

The other change is to parse a version file if it exists.

charmdir.go
```diff
@@ -246,21 +254,12 @@ func resolveSymlinkedRoot(rootPath string) (string, error) {
 // ArchiveTo creates a charm file from the charm expanded in dir.
 // By convention a charm archive should have a ".charm" suffix.
 func (dir *CharmDir) ArchiveTo(w io.Writer) error {
-	versionString, vcsType, err := dir.MaybeGenerateVersionString()
```
SimonRichardson:
```go
// If the dir.version is empty then go and generate the version string
if dir.version == "" {
    // Do this here as well - dir.MaybeGenerateVersionString()
```

charmdir.go
nammn: thanks for the err catch!
I would like to stay consistent with the others how they handle file parsing. 
I actually want to make sure that the version file is correct and otherwise we log so (?)

charmdir.go
nammn: good questions... My ide did show that pretty hard :D

mocks/logger_mock.go
```diff
@@ -0,0 +1,21 @@
+package mocks
```
SimonRichardson: Let's not call this a mock logger or put it in the mock folder, we don't want to confuse existing mocking setup. I'd rather we call this `NopLogger`, as it performs no operations. You could stick this in the same file as the interface, it's just an implementation of the interface then.

charmdir.go
```diff
@@ -425,9 +459,22 @@ func (dir *CharmDir) MaybeGenerateVersionString() (string, string, error) {
 		cmdArgs = []string{"bzr", "version-info"}
 	} else if _, err = os.Stat(filepath.Join(dir.Path, ".git")); err == nil {
 		cmdArgs = []string{"git", "describe", "--dirty", "--always"}
+	} else if file, err := os.Open(dir.join("version")); err == nil {
+		logger.Debugf("charm is not in version control, but uses a version file")
```
SimonRichardson: I guess one thing to note is that if you move from version to scm or vice-versa you could end up with issues around tracking the version, but I guess they're probably holding it wrong, as there isn't a way to know if it once was with the prior revision control.

nammn: I think one could add additional thoughts into it. 
But this would be outside of the scope of this PR.

Thanks for the review though!

nammn:
$$merge$$

PR #296, opened 2019-10-24
logging: invert logging place

nammn:
inverts logging place, only on archiveTo as a charm can be read way more times in a lifetime than archived. Additionally adding the path helps a lot and reducing the level to Info

This should lead to the following:

- less logging in general
- less confusing messages
- no warning or higher unless it needs to

### DIFF
--- a/charmdir.go
+++ b/charmdir.go
@@ -120,11 +120,9 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
 			return nil, errors.New("invalid revision file")
 		}
 	}
-	version, _, err := dir.MaybeGenerateVersionString(logger)
-	if err != nil {
-		// We don't want to stop, even if the version cannot be generated
-		logger.Errorf("unexpected problem trying to generate version string: %q", err)
-	}
+
+	var dummyLogger = NopLogger{}
+	version, _, err := dir.MaybeGenerateVersionString(dummyLogger)
 	dir.version = version
 
 	file, err = os.Open(dir.join("lxd-profile.yaml"))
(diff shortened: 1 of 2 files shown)


charmdir.go
```diff
@@ -120,11 +120,9 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
...
 		}
 	}
-	version, _, err := dir.MaybeGenerateVersionString(logger)
-	if err != nil {
-		// We don't want to stop, even if the version cannot be generated
-		logger.Errorf("unexpected problem trying to generate version string: %q", err)
-	}
+
+	var dummyLogger = NopLogger{}
+	version, _, err := dir.MaybeGenerateVersionString(dummyLogger)
```
nammn: I suspected that the past change here can lead to a high volume of logging, as a charm can be read more than it will be "archived"

charmdir.go
```diff
@@ -460,21 +467,21 @@ func (dir *CharmDir) MaybeGenerateVersionString(logger Logger) (string, string,
...
+		logger.Debugf("charm is not in version control, but uses a version file, charm path %q", dir.Path)
 		var versionNumber string
 		n, err := fmt.Fscan(file, &versionNumber)
 		if n != 1 {
-			return "", "versionFile", errors.Errorf("invalid version file")
+			return "", versionFileVersionType, errors.Errorf("invalid version file, charm path: %q", dir.Path)
 		}
 		if err != nil {
-			return "", "versionFile", err
+			return "", versionFileVersionType, err
```
nammn: I was thinking about taking all of those "vcstype" out into global constants. Thus we could reuse them e.g. in tests and do not have to use "magical" strings. But this would warrant additional changes I wasn't ready to commit

nammn:
$$merge$$

PR #297, opened 2019-10-24
version file check: example how version check can be done

nammn:
This PR is only used as a discussion point to changing the way we find the version of charms.

Unsure about the way we do the err handling.
Still needs to adapt the unit tests
uses cmd quite a lot (I don't like)

## Problem Current Implementation
- Does not take into consideration if the charm is a subdir of a versioned charm

## Possible Solution 1
- Try each vcs
- solves the above problem
### Disadvantage
- Makes more sophisticated err handling difficult. Let's say a `git describe ...` fails for some reason the code just thinks it is not related to git and "tries" the next one
   - We can either log on every fail case -> convolute Log 
   - We do not log -> missing information for case it fails but actually is under git

## Possible Solution 2
- Try each vcs
- for each vcs (especially git) have a own func which handles err and as well as does a check whether it exists or not

## Unsure
- each implementation calls cmd in worst case get's calls multiple times

### DIFF
--- a/charmdir.go
+++ b/charmdir.go
@@ -6,7 +6,6 @@ package charm
 import (
 	"archive/zip"
 	"fmt"
-	"github.com/juju/errors"
 	"io"
 	"os"
 	"os/exec"
@@ -14,6 +13,8 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+
+	"github.com/juju/errors"
 )
 
 // defaultJujuIgnore contains jujuignore directives for excluding VCS- and
@@ -121,7 +122,7 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
 	}
 
 	var dummyLogger = NopLogger{}
-	version, _, err := dir.MaybeGenerateVersionString(dummyLogger)
+	version, _, _ := dir.MaybeGenerateVersionString(dummyLogger)
 	dir.version = version
 
 	file, err = os.Open(dir.join("lxd-profile.yaml"))
@@ -263,7 +264,7 @@ func (dir *CharmDir) ArchiveTo(w io.Writer) error {
 	dir.version, _, err = dir.MaybeGenerateVersionString(logger)
 	if err != nil {
 		// We don't want to stop, even if the version cannot be generated
-		logger.Errorf("unexpected problem trying to generate version string: %q", err)
+		logger.Warningf("%v", err)
 	}
 
 	return writeArchive(w, dir.Path, dir.revision, dir.version, dir.Meta().Hooks(), ignoreRules)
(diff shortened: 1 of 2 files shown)


charmdir.go
```diff
@@ -454,19 +453,69 @@ func (m NopLogger) Infof(message string, args ...interface{}) {
...
+	args := []string{"rev-parse", "--is-inside-work-tree"}
+	cmd := exec.Command("git", args...)
+	if _, err := cmd.Output(); err == nil {
+		return true
+	}
+	return false
+}
+
+func handleErrGit(err error, vcsType string, charmPath string) {
+	logger.Debugf("%q version string generation failed : %v\nThis means that the charm version won't show in juju status. Charm path %q", vcsType, err, charmPath)
```
SimonRichardson: I'm unsure what the need for `vcsType` here is, as you know you're handling `git`.

charmdir.go
```diff
@@ -454,19 +453,69 @@ func (m NopLogger) Infof(message string, args ...interface{}) {
 
 }
 
+type vcsCMD struct {
+	vcsType       string
+	args          []string
+	usesTypeCheck func() bool
+	errHandler    func(err error, vcsType string, path string)
+	charm         *CharmDir
```
SimonRichardson: I don't think `charm *CharmDir` is required here.

charmdir.go
```diff
@@ -454,19 +453,69 @@ func (m NopLogger) Infof(message string, args ...interface{}) {
...
-	if _, err := os.Stat(filepath.Join(dir.Path, ".hg")); err == nil {
-		cmdArgs = []string{"hg", "id", "-n"}
-	} else if _, err = os.Stat(filepath.Join(dir.Path, ".bzr")); err == nil {
-		cmdArgs = []string{"bzr", "version-info"}
-	} else if _, err = os.Stat(filepath.Join(dir.Path, ".git")); err == nil {
-		cmdArgs = []string{"git", "describe", "--dirty", "--always"}
-	} else if file, err := os.Open(dir.join("version")); err == nil {
+	mercurialStrategy := vcsCMD{"hg", []string{"id", "-n"}, notImplementedUsesCheck, notImplementedErrHandler, dir}
+	bazaarStrategy := vcsCMD{"bzr", []string{"version-info"}, notImplementedUsesCheck, notImplementedErrHandler, dir}
+	gitStrategy := vcsCMD{"git", []string{"describe", "--dirty", "--always"}, usesGit, handleErrGit, dir}
```
SimonRichardson: Drop `, dir}` - it's not required, as it will be the same for all strategies.

charmdir.go
nammn: yeah you are right, had this initially in there and somehow lost it :D

charmdir.go
```diff
@@ -454,19 +453,84 @@ func (m NopLogger) Infof(message string, args ...interface{}) {
...
+	vcsStrategies["hg"] = mercurialStrategy
+	vcsStrategies["git"] = gitStrategy
+	vcsStrategies["bzr"] = bazaarStrategy
+
+	// Nowadays most vcs used are git, we want to make sure that git is the first one we test
+	vcsOrder := [...]string{"git", "hg", "bzr"}
+
+	for _, vcsType := range vcsOrder {
+		vcsCmd := vcsStrategies[vcsType]
+		if !vcsCmd.usesTypeCheck(dir.Path) {
```
SimonRichardson: Can we add a comment that if all vcs strategies fail, then it will fall back to the version check below the for loop.

nammn:
$$merge$$

PR #300, opened 2019-12-11
1855777 invalid bundle applications

tlmiller:
Adds test and fix for nil applications in verify bundle, updated go mode so it will now compile the project.


bundledata_test.go
```diff
@@ -1449,6 +1449,34 @@ func (*bundleDataSuite) TestParsePlacement(c *gc.C) {
...
+    application2:
+`,
+	}
+
+	for _, d := range tstDatas {
+		bd, err := charm.ReadBundleData(strings.NewReader(d))
+		c.Assert(err, gc.IsNil)
+
+		err = bd.Verify(nil, nil, nil)
+		c.Assert(err, gc.NotNil)
```
hpidcock: c.Assert(err, gc.ErrorMatches, `bundle application for key .+ is undefined`)

tlmiller: Done & Done

tlmiller:
$$merge$$

PR #302, opened 2020-01-13
charm logging: add abspath

nammn:
- adds abspath output, as is could be relative, which makes it hard for debugging purposes
- some unit tests with it

### DIFF
--- a/charmdir.go
+++ b/charmdir.go
@@ -534,6 +534,15 @@ func (dir *CharmDir) MaybeGenerateVersionString(logger Logger) (string, string, 
 	vcsOrder := [...]string{"git", "hg", "bzr"}
 	cmdWaitTime := 2 * time.Second
 
+	absPath := dir.Path
+	if !filepath.IsAbs(absPath) {
+		var err error
+		absPath, err = filepath.Abs(dir.Path)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


charmdir.go
```diff
@@ -534,6 +534,15 @@ func (dir *CharmDir) MaybeGenerateVersionString(logger Logger) (string, string,
 	vcsOrder := [...]string{"git", "hg", "bzr"}
 	cmdWaitTime := 2 * time.Second
 
+	absPath := dir.Path
+	if !filepath.IsAbs(absPath) {
+		var err error
+		absPath, err = filepath.Abs(dir.Path)
+		if err != nil {
+			logger.Warningf("failed resolving relative path %q", dir.Path)
```
nammn: We don't want to fail here, as this would change the long-running behavior we had until now imo

achilleasa: So, what happens if this fails and you fall through? Won't it cause issues with the code below? I think that absPath will be set to `""`

nammn: Hmm you are right, i suppose fallback to path would be the way to have no regression?

nammn: returning the error now.

nammn:
$$merge$$ thanks for the rev!

PR #303, opened 2020-01-24
We no longer support functions so revert back to actions

wallyworld:
Remove the code to load either actions or functions and just support actions once again.


wallyworld:
$$merge$$

PR #305, opened 2020-03-13
Introduce relation-created hook

achilleasa:
This new hook is the counterpart of the relation-broken hook.


achilleasa:
$$merge$$

PR #306, opened 2020-03-25
Add k8s deployment mode for workload vs operator

wallyworld:
k8s charms can act as operators - they do not ask Juju to deploy something like a database workload, but instead directly create resources in the cluster.

This PR adds "mode" the deployment metadata to enable a k8s charm to say that it is acting as an "operator" (vs a "workload").
Juju will manage such charms differently to workload charms.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -158,6 +158,14 @@ const (
 	DeploymentDaemon    DeploymentType = "daemon"
 )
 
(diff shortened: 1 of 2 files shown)


tlm:
lgtm

wallyworld:
$$merge$$

PR #307, opened 2020-03-25
Ensure deployment block is correctly parsed with mode

wallyworld:
Remove a bogus check for empty deployment block. It will be picked up by the schema parser.
Also add a new test for deployment mode.


wallyworld:
$$merge$$

PR #308, opened 2020-03-26
Remove checks for non statefull depoyment type with storage;

ycliuhw:
Remove checks for non statefull depoyment type with storage;


wallyworld:
$$merge$$

PR #309, opened 2020-04-15
Remove injected default limit of 1 for peer/requirer relations

achilleasa:
While working on a fix for https://bugs.launchpad.net/juju/+bug/1869840, I realized that the charm.v6 package is using an odd schema-coercion mechanism that ends up injecting a default limit of **1** for `peer` and `requires` relations (provides is subject to no limit).

These defaults cause problems when attempting to actually enforce the limit as there are several valid use cases where forcing such a default is wrong. For example, a metrics collector charm defines a required relation and should be relate-able to multiple applications as a subordinate.

To this end, this PR removes the above limits and aligns our implementation with the [docs](https://discourse.juju.is/t/implementing-relations/1051):

```
limit is ignored by Juju, but if present should be a positive integer N indicating that the charm is not designed to use this interface in more than N relations at once.
```

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -610,11 +610,7 @@ type marshaledRelation Relation
 
 func (r marshaledRelation) MarshalYAML() (interface{}, error) {
 	// See calls to ifaceExpander in charmSchema.
-	noLimit := 1
-	if r.Role == RoleProvider {
-		noLimit = 0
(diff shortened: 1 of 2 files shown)


achilleasa:
$$merge$$

PR #310, opened 2020-06-25
Allow for better error messages

SimonRichardson:
When attempting to decode the charm it is very useful to bubble the
information about which file has broken the charm, so the operator can
attempt to fix it quickly and easily.

Fixes: https://bugs.launchpad.net/juju/+bug/1871711

### DIFF
--- a/charmdir.go
+++ b/charmdir.go
@@ -70,24 +70,24 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
 	dir = &CharmDir{Path: path}
 	file, err := os.Open(dir.join("metadata.yaml"))
 	if err != nil {
-		return nil, err
+		return nil, errors.Annotatef(err, `issue reading "metadata.yaml" file`)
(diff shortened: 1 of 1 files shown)


SimonRichardson:
$$merge$$

charmdir.go
```diff
@@ -70,24 +70,24 @@ func ReadCharmDir(path string) (dir *CharmDir, err error) {
 	dir = &CharmDir{Path: path}
 	file, err := os.Open(dir.join("metadata.yaml"))
 	if err != nil {
-		return nil, err
+		return nil, errors.Annotatef(err, `issue reading "metadata.yaml" file`)
```
manadart, line 73: I prefer these annotations to always be only what we were doing at the time. We'll know there was an issue/error based on the log level we choose for it, making "error doing x" or "issue with y" redundant.

So it would be `reading "metadata.yaml" file`

PR #311, opened 2020-07-09
Charm origin

SimonRichardson:
With the addition of different stores from which you can install a charm
from, we need to correctly model where a charm originates from. The
changes here allow just the origin source for now, I can see in the
future that more fields will want to be added.

The possible sources that we need to model are charmhub and everything
else. With this change though we can give feedback to the operator about
where a charm originates from to help them understand why a charm might
not be updated without switching or updating the path.

Without correctly modelling this we have to inspect the URL of a charm
and that's just more tricky. If we know when we deploy a charm what the
source was, we should store that.

---

The code is really simple, it's an optional addition to a charm, so
libraries using this won't have to do anything.

### DIFF
--- a/charmarchive.go
+++ b/charmarchive.go
@@ -29,6 +29,7 @@ type CharmArchive struct {
 	metrics    *Metrics
 	actions    *Actions
 	lxdProfile *LXDProfile
+	origin     *Origin
 	revision   int
 	version    string
 }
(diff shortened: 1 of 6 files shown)


SimonRichardson:
It doesn't look like this is the best approach. I may re-open this, but closing this for now.

PR #312, opened 2020-08-10
Expose rewriting of charmhub URLs

SimonRichardson:
In theory charmhub URLs should never be written into the database as the
version 2 of the charmhub URL. Yet, we should do our utmost to ensure
that is the case. By exposing the re-write capabilities of our charm URL
function, we can re-use that with in Juju directly and ensure that all
charm URLs are of the correct fashion.

----

Essentially the v2 charm URL re-writes from a URL that has a schema (not
RFC3986 URL Schema) and re-arranges it into a v1 charm URL.

Example:
 - `http://<domain>/<name>` becomes `cs:<name>`

So if any "store" implements this schema, it can be re-written into the v1 charm
URL.

> `https://<domain>/u/<user>/<name>/<series>/<revision>`

Whilst clever, it conflates the charm URL parsing code and causes issues when
migrating away from the charm store.

I guess the idea was to allow people to copy URLs directly from a browser into
Juju CLI directly. Unsure how much this is supported, considering most if not
all tutorials/documentation refers to v1 charm URL extensively.

### DIFF
--- a/url.go
+++ b/url.go
@@ -116,14 +116,6 @@ func MustParseURL(url string) *URL {
 //    https://jujucharms.com/u/user/name/series
 //    https://jujucharms.com/u/user/name/revision
 //    https://jujucharms.com/u/user/name/series/revision
-//    https://jujucharms.com/channel/name
-//    https://jujucharms.com/channel/name/series
-//    https://jujucharms.com/channel/name/revision
-//    https://jujucharms.com/channel/name/series/revision
-//    https://jujucharms.com/u/user/channel/name
-//    https://jujucharms.com/u/user/channel/name/series
-//    https://jujucharms.com/u/user/channel/name/revision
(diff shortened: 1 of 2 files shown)


url.go
```diff
@@ -400,3 +334,89 @@ func Quote(unsafe string) string {
...
+		return "", errors.Errorf("charm or bundle URL %q has unrecognized parts", url)
+	}
+	httpURL, err := parseHTTPURL(u)
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+	return httpURL.String(), nil
+}
+
+func parseHTTPURL(url *gourl.URL) (*URL, error) {
```
SimonRichardson, line 380: Renamed to highlight what the method actually does.

SimonRichardson:
$$merge$$

PR #313, opened 2020-08-11
Allow charmhub url parsing

SimonRichardson:
The following updates the charm URL code to allow the parsing of
charmhub prefix schema. This is a breaking change and will require a v8
of the charm package!

Essentially all charm URLs will require a schema prefix in every case.
Juju or other consumers will have to lift the raw url into the right
schema so that you get the right url.

---

## Without a revision

Charmhub URLs can be used without a revision, and will just contain the application name.

```
cs:wordpress
```

## With the revision

```
cs:wordpress-1
```

### DIFF
--- a/bundle_test.go
+++ b/bundle_test.go
@@ -68,21 +68,21 @@ func checkWordpressBundle(c *gc.C, b charm.Bundle, path string) {
 	mysqlCharm := readCharmDir(c, "mysql")
 
 	bd := b.Data()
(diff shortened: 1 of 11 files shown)


SimonRichardson:
https://trello.com/c/TQlUlxU7/2843-handle-charmhub-in-the-charm-url

SimonRichardson:
$$merge$$

PR #315, opened 2020-09-04
Augment bundle data application model with fields for app expose params

achilleasa:
This PR adds the following (optional) fields to the application specification metadata for bundles:

- exposed-endpoints
- expose-to-spaces
- expose-to-cidrs

These fields will be used by the work for exposing individual application endpoints and/or for allowing access from specific spaces/CIDRs to exposed applications.

PR #316, opened 2020-09-04
Only allow expose-to-{spaces,CIDRs} fields to be present in overlays

achilleasa:
This is a followup to #315 to ensure that the `expose-to-spaces` and `expose-to-cidrs` fields can only be specified as part of an overlay whereas the `exposed-endpoints` field can be specified in either the base or the overlay part of a bundle.

The reasoning behind this change is both the space names and any CIDRs that you want to allow access to exposed applications are deployment-specific. As such, they should be provided via an overlay similar to the ACL bits for offers.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -295,11 +295,11 @@ type ApplicationSpec struct {
 
 	// ExposeToSpaces contains a list of spaces that should be able to
 	// access the application ports if the application is exposed.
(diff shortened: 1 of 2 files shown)


manadart (approved):

achilleasa:
$$merge$$

PR #317, opened 2020-09-07
Support per-endpoint expose params for space/CIDRs

achilleasa:
This PR revises the work from #315 and #316 to  allow bundle authors 
to specify the list of spaces and/or CIDRs that can access the opened 
ports for exposed applications on a per-endpoint basis.

Given that the exposed-endpoint details are deployment specific, they
can only be specified as part of an overlay document. In addition, the
exposed-endpoint field is not allowed to be specified in tandem with the
'exposed: true' flag.

This is intentional: imagine a scenario where an operator exports a
bundle from a 2.9 controller where only a subset of an application's
endpoints have been exposed (e.g. to be accessed via a jumpbox) and
attempts to deploy the bundle to a 2.8 or older controller. If we
allowed both the "exposed: true" and the "exposed-endpoint" fields to be
present in the bundle, the 2.8 controller (which does not recognize the
"exposed-endpoint" field) would incorrectly assume that *all* endpoints
are to be exposed (due to "exposed: true"), thus opening a security
hole.

By making these two fields mutually exclusive, we mitigate this security
risk. In fact, as the export-bundle command will skip the "exposed: true"
flag and only emit the "exposed-endpoint" field, the bundle will not
cause the application to be exposed in older controllers and instead,
require the operator to audit the bundle and manually export the
application where needed.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -287,19 +287,15 @@ type ApplicationSpec struct {
 	// Expose holds whether the application must be exposed.
 	Expose bool `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 
-	// ExposedEndpoints stores a subset of the application endpoints that
-	// are used to select the set of open ports that should be accessible
-	// if the application is exposed. An empty value indicates that all
-	// open ports should be made accessible.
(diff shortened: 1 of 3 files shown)


SimonRichardson (approved):
This is much better, as we group everything into a map. 
Easier to see how things are related.

achilleasa:
$$merge$$

PR #319, opened 2020-09-09
Force the charmhub naming for URLs

SimonRichardson:
The following forces non-prefix charm URLs to fallback to charmhub
schemas. This is a much bigger breaking change than the previous
changes. This states that anything withouth a prefix is a charmhub,
where previously that was stated as a charmstore charm URL.

Additionally we drop the prefix of a charm URL during the string
representation of a URL. To get the previous same logic, calling the
following, URL#FullPath() will do it.

This has been tricky as we want to preserve as much of the old way as
possible, but we also don't want to keep things around for no reason.

### DIFF
--- a/url.go
+++ b/url.go
@@ -134,8 +134,8 @@ func ValidateName(name string) error {
 
 // WithRevision returns a URL equivalent to url but with Revision set
 // to revision.
-func (url *URL) WithRevision(revision int) *URL {
-	urlCopy := *url
+func (u *URL) WithRevision(revision int) *URL {
+	urlCopy := *u
 	urlCopy.Revision = revision
 	return &urlCopy
 }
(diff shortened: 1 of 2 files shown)


url.go
```diff
@@ -251,28 +251,41 @@ func parseV1URL(url *gourl.URL, originalURL string) (*URL, error) {
...
+	if u.Series != "" {
+		parts = append(parts, u.Series)
 	}
-	if r.Revision >= 0 {
-		parts = append(parts, fmt.Sprintf("%s-%d", r.Name, r.Revision))
+	if u.Revision >= 0 {
+		parts = append(parts, fmt.Sprintf("%s-%d", u.Name, u.Revision))
 	} else {
-		parts = append(parts, r.Name)
+		parts = append(parts, u.Name)
```
achilleasa, line 265: Curious as to how the store deals with URLs that have variable arguments... Does it pattern match on ~ for user and %s-%d for revision?

url.go
```diff
@@ -251,28 +251,41 @@ func parseV1URL(url *gourl.URL, originalURL string) (*URL, error) {
...
+
+// Path returns the path of the URL without the schema.
+func (u *URL) Path() string {
+	return u.path()
 }
 
-func (u URL) String() string {
-	return fmt.Sprintf("%s:%s", u.Schema, u.Path())
+// String returns the string representation of the URL.
+// To keep backwards with older schema versions (CharmStore), we output the
```
achilleasa, line 281: s/backwards/backwards compatibility/?

url.go
```diff
@@ -251,28 +251,41 @@ func parseV1URL(url *gourl.URL, originalURL string) (*URL, error) {
...
+// FullPath of the URL. For new CharmHub integrations, we only want the Path.
+// That way we can hide the schema to the user.
```
achilleasa, line 283: s/to the/from the/?

achilleasa (approved):

url.go
SimonRichardson: As the charmstore uses the same code as this, or at least it did! Then it parsed it to extrapolate that same information that we had it. So in a sense, the URL would be isomorphic between juju and the charmstore. As we never pass the URL directly to the charmhub any longer, we are abusing the URL to just store the information we require, namely `<schema>:<name>-<revision>`. It has valid useful information in the URL and more importantly, it keeps Juju happy so that we don't have to re-write all the works that watch for the URL to change.

The longer term, this should be renamed to something other than URL, but that would be a mountain of work.

SimonRichardson:
$$merge$$

url.go
achilleasa: Really? Guessing the meaning of args when parsing URLs is just a disaster waiting to happen... Any idea why these are not encoded as query string parameters in the first place?

SimonRichardson: You see that would have been sane!

`<schema>://<username>@<name>?<revision>=&<series>=`

PR #322, opened 2020-11-17
Update to github.com/juju/os/v2

benhoyt:
This is so that juju/juju only pulls in one version of juju/os now that we're using v2 there. Related PR: https://github.com/juju/juju/pull/12329


wallyworld (approved):
It's not just so juju only pulls in one version, it's also to ensure the same semantics for juju and charm repos wrt series handling

PR #324, opened 2020-11-18
Include ch schema in String().

hmlanigan:
If we do not include the 'ch' schema in the String(), a *charm.URL is
not saved to the juju db with the ch schema.  It appears that mongo is
saving via the String().  This may create issues down the road when a
new charm schema is created.

User facing messages in juju have been cleaned up so that
charm.URL.String() is no longer directly printed.

### DIFF
--- a/url.go
+++ b/url.go
@@ -278,14 +278,7 @@ func (u *URL) Path() string {
 }
 
 // String returns the string representation of the URL.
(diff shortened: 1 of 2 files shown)


SimonRichardson (approved):

SimonRichardson:
$$merge$$

PR #326, opened 2020-12-18
Remove legacy services

SimonRichardson:
Services have been time and memorial, unfortunately, they were
deprecated in version 2.x (2016 timeframe). I think authors of charms
have had enough time to move over to applications.

This removes services, yet we have to keep the horrid hack of
normalizing the data when we unmarshal. It's wrong and shouldn't be in
this code, it should be a two-step process and remove the magic that
causes us issues in the long run.

---

This is a breaking change and we'll bump the version appropriately.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -23,101 +23,6 @@ import (
 
 const kubernetes = "kubernetes"
 
-type noMethodsBundleData BundleData
-
-type legacyBundleData struct {
-	noMethodsBundleData `bson:",inline" yaml:",inline" json:",inline"`
-
-	// LegacyServices holds application entries for older bundle files
-	// that have not been migrated to use the new "application" terminology.
-	LegacyServices map[string]*ApplicationSpec `json:"services" yaml:"services" bson:"services"`
-}
-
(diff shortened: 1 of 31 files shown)


bundledata.go
```diff
@@ -336,7 +230,96 @@ type ApplicationSpec struct {
...
+type maskedBundleData BundleData
+
+type bundleData struct {
+	maskedBundleData `bson:",inline" yaml:",inline" json:",inline"`
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (bd *BundleData) UnmarshalJSON(b []byte) error {
+	var in bundleData
+	if err := json.Unmarshal(b, &in); err != nil {
```
achilleasa, line 253: I think we should detect the presence of services in bundle data and return an error to indicate they are not supported.

achilleasa (approved):

bundledata.go
SimonRichardson: Maybe... I'm unsure about this.

SimonRichardson:
$$merge$$

PR #329, opened 2021-01-14
Allow architecture and series in identifier URLs

SimonRichardson:
Identifier URLs (charmhub URLs) are required to now identify themselves
with ability to adorn themselves with architectures and series. This is
so the uniquness of a URL within Juju is more than just the name and
revision. It allows more fine grained constraints about what a charm URL
is.

It does have the consequence that any other repsonsibilities outside of
the constrainted URL domain is now lost. If you want to use the charm
URL for a differing architecture becomes unwise or even impossible.

### DIFF
--- a/url.go
+++ b/url.go
@@ -76,14 +76,16 @@ type Location interface {
 //     ch:wordpress
 //
 type URL struct {
-	Schema   string // "cs", "ch" or "local".
-	User     string // "joe".
-	Name     string // "wordpress".
-	Revision int    // -1 if unset, N otherwise.
-	Series   string // "precise" or "" if unset; "bundle" if it's a bundle.
+	Schema       string // "cs", "ch" or "local".
+	User         string // "joe".
+	Name         string // "wordpress".
+	Revision     int    // -1 if unset, N otherwise.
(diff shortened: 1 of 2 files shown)


url.go
```diff
@@ -76,14 +76,16 @@ type Location interface {
...
+	Schema       string // "cs", "ch" or "local".
+	User         string // "joe".
+	Name         string // "wordpress".
+	Revision     int    // -1 if unset, N otherwise.
+	Series       string // "precise" or "" if unset; "bundle" if it's a bundle.
+	Architecture string // "amd64" or "" if unset for charmstore (v1) URLs.
 }
 
 var (
+	validArch   = regexp.MustCompile("^[a-z]+([a-z0-9]+)?$")
```
SimonRichardson, line 88: Is there any arches out there that start with a number?

hmlanigan (approved):
Using in juju with bundlechanges pr 78, I was able to deploy 2 units of different applications using the same charm. There were series differences and architecture differences between the setup of each application.

SimonRichardson:
$$merge$$

PR #332, opened 2021-02-04
Application short syntax

SimonRichardson:
The following implements application short syntax as the following:

    applications:
        mysql: {}

There are valid reasons for doing this. Initially, it was thought that ~
could be used to indicate the short syntax, but ~ already indicates
null. This is fine but causes all sorts of hell when integrating with
overlays. ~ becomes a way to delete charms and that makes it confusing
and very hard to work out if we should use the short syntax or delete a
charm.

The best way forward is to just sidestep this completely and use the
empty object type. It's not as short, but it's very clear from the code
how to move forward with it.

Additionally, this also adds the default number of units to 1, which also
handles the Scale and NumUnits conversion. Essentially Scale can just be
ignored once it's been normalized.

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -65,6 +65,10 @@ type BundleData struct {
 
 	// Short paragraph explaining what the bundle is useful for.
 	Description string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+
+	// alteredApplications reference applications that have been altered in
+	// normalization step.
+	alteredApplications map[string]struct{}
 }
 
 // SaasSpec represents a single software as a service (SAAS) node.
@@ -230,6 +234,9 @@ type ApplicationSpec struct {
 	RequiresTrust bool `bson:"trust,omitempty" json:"trust,omitempty" yaml:"trust,omitempty"`
 }
 
+// Empty works out if an ApplicationSpec is empty or not.
+// We can't use an equality check here because of maps with in the type and
+// using reflect.IsZero returns false because sometimes maps can be initialized.
 func (s *ApplicationSpec) Empty() bool {
 	return s.Charm == "" && s.Channel == "" && s.Series == "" &&
 		len(s.Resources) == 0 && s.NumUnits == nil && s.Scale_ == nil &&
@@ -305,9 +312,19 @@ func (bd *BundleData) normalizeData() error {
 		if app.Empty() {
 			app.Charm = appName
 			app.NumUnits = newInt(1)
+
+			if bd.alteredApplications == nil {
+				bd.alteredApplications = make(map[string]struct{})
+			}
+			bd.alteredApplications[appName] = struct{}{}
 		}
 		if app.Scale_ == nil && app.NumUnits == nil {
 			app.NumUnits = newInt(1)
+
+			if bd.alteredApplications == nil {
+				bd.alteredApplications = make(map[string]struct{})
+			}
+			bd.alteredApplications[appName] = struct{}{}
 		}
 		// Kubernetes bundles use "scale" instead of "num_units".
 		if (app.Scale_ != nil && *app.Scale_ > 0) && (app.NumUnits != nil && *app.NumUnits > 0) {
@@ -421,7 +438,6 @@ type bundleDataVerifier struct {
 	charms map[string]Charm
 
 	errors            []error
-	warnings          map[string][]string
 	verifyConstraints func(c string) error
 	verifyStorage     func(s string) error
 	verifyDevices     func(s string) error
@@ -433,14 +449,6 @@ func (verifier *bundleDataVerifier) addErrorf(f string, a ...interface{}) {
 
 func (verifier *bundleDataVerifier) addError(err error) {
 	verifier.errors = append(verifier.errors, err)
-}
-
-func (verifier *bundleDataVerifier) addWarningf(name, f string, a ...interface{}) {
-	verifier.addWarning(name, fmt.Sprintf(f, a...))
-}
-
-func (verifier *bundleDataVerifier) addWarning(name, err string) {
-	verifier.warnings[name] = append(verifier.warnings[name], err)
 }
 
 func (verifier *bundleDataVerifier) err() error {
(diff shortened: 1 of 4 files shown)


bundledata.go
```diff
@@ -230,6 +234,20 @@ type ApplicationSpec struct {
...
+func (s *ApplicationSpec) Empty() bool {
+	return s.Charm == "" && s.Channel == "" && s.Series == "" &&
+		len(s.Resources) == 0 && s.NumUnits == nil && s.Scale_ == nil &&
+		len(s.To) == 0 && s.Placement_ == "" && !s.Expose &&
+		len(s.ExposedEndpoints) == 0 && len(s.Options) == 0 &&
+		len(s.Annotations) == 0 && s.Constraints == "" &&
+		len(s.Storage) == 0 && len(s.Devices) == 0 &&
+		len(s.EndpointBindings) == 0 && len(s.Offers) == 0 &&
+		s.Plan == "" && !s.RequiresTrust
+}
```
petevg, line 249: My Python brain wants this to be constructed something like:

```
    for key in s.keys:
        if not_empty(key):
            return True
    # if everything is empty, return empty
    return False
```

In Python, not_empty would be a function that could take anything as an arg, which branched based on type. In Erlang, not_empty would be a series of functions, pattern matched based on the type of the arg. I'm not sure what the best practice is in Go. But it definitely feels wrong to me to see a big chunk of text that needs to be hand edited if the bundle spec ever changes. (Worse, if it's not hand edited, it can return a false negative if a new key, and only the new key, is empty.)

bundledata.go
```diff
@@ -291,13 +309,30 @@ func (bd *BundleData) normalizeData() error {
...
 		// Kubernetes bundles use "scale" instead of "num_units".
-		if app.Scale_ > 0 && app.NumUnits > 0 {
+		if (app.Scale_ != nil && *app.Scale_ > 0) && (app.NumUnits != nil && *app.NumUnits > 0) {
 			return fmt.Errorf("cannot specify both scale and num_units for application %q", appName)
 		}
-		if app.Scale_ > 0 && app.NumUnits == 0 {
+		if (app.Scale_ != nil && *app.Scale_ > 0) && (app.NumUnits == nil || *app.NumUnits == 0) {
 			app.NumUnits = app.Scale_
-			app.Scale_ = 0
+			app.Scale_ = nil
```
petevg, line 335: I like this change to how scale is handled. Much nicer!

bundledata.go
SimonRichardson: So you can do this automatically via the `reflect` package, but we tend to stay away from that, as it generates more complex code than you can easily reason about.

We could use `reflect.ValueOf(s).IsZero()`, but that doesn't work either, as that expects everything to be nil or the default value and that isn't true either. It could be that some fields have empty slices/maps that have been created during (de)serialization, but aren't nil, so we consider it empty if those are of length zero.

This becomes painful when you consider that this logic also has to work for overlays and then we have to ensure that every value can be part of an overlay and cause changes to the base bundle. That means that we can't just check for one value either and hopefully that'll give us enough of a sign for emptiness.

I can move back to reflect so that you can get things for free, but when somebody adds a type to the bundle and it isn't covered by all the reflect kinds (`map[interface{}]interface{}` and no `map[string]interface{}`) then you're in the same spot, but now you have the burden of understanding the empty function with the complex reflect shenanigans.

Lastly, I could use `json.Marshall` to a set of bytes and see if the length is expected, but then you need to ensure `omitempty` is in the right place. 

It's kind of annoying, but this is golang. 
Read many, write once.

bundledata.go
```diff
@@ -113,11 +117,11 @@ type ApplicationSpec struct {
 	// For a subordinate application, this actually represents
```
hmlanigan, line 117: This comment needs updating.  Before the prior work (2015) to make NumUnits not a pointer it was:

// For a subordinate service, this must be nil;
// otherwise a nil value represents the default
// of a single unit.

bundledata.go
```diff
@@ -291,13 +309,30 @@ func (bd *BundleData) normalizeData() error {
 		if app == nil {
 			continue
 		}
+		if app.Empty() {
+			app.Charm = appName
+			app.NumUnits = newInt(1)
```
hmlanigan, line 314: subordinates can't be deployed with a count.

bundledata_test.go
```diff
@@ -1442,6 +1445,63 @@ applications:
 	}
 }
 
+// Tests that empty/nil applications cause an error
+func (*bundleDataSuite) TestApplicationShortSyntax(c *gc.C) {
```
hmlanigan, line 1449: Are there any unit tests to cover short syntax and subordinates?

bundledata.go
SimonRichardson: So there is indeed a problem. Essentially we don't have enough information at the time of unpacking to determine if a charm is a subordinate. As we only have a charm URL and they do not identify a subordinate and only the charm metadata expresses this option, we are stuck.

The way I implemented it was that subordinates will never be allowed a short syntax, there isn't anything we can do about it with the current setup.

As this package is a modeling library where we don't actually confirm the existence of a given charm, just values appear to be correct for deployment, we're in a bit of a sticky situation.

----

The previous approach was to use `VerifyWithCharms` which requires that you know all the charms upfront. This isn't the case at all with Juju. So that was a fallacy.

----

In reality, we probably want to expose a `NumUnits` method that we use to get the correct number of default units and use that throughout Juju. We can check the `alteredApplications` method to see if the subordinate was using the short syntax originally.

SimonRichardson: I don't believe Juju itself uses NumUnits from the bundle code directly and actually we might only be using from bundlechanges, we could use the `NumUnits(subordinate)` to gain that information.

SimonRichardson:
@jameinel We can't use empty or `~` (null and undefined) syntax as that implies deletion of a charm for overlays, which leads to a confusing experience. A user can specify an empty charm for only base bundles, but that same behavior leads to destructive changes when you do the same in overlays. That's firing alarm bells for me.

I have a feeling we're pushing the limits of the expressiveness of the YAML language for a bundle.

SimonRichardson:
Closing this one for now, turns out overlays can also specify with `{}` and deletes a charm from the base bundle.

PR #336, opened 2021-03-05
Prevent series and systems

SimonRichardson:
Metadata V2 should not be allowed into charmhub without the appropriate
manifest.yaml. To prevent this from becoming the norm where V1 and V2 are
intermixed and therefore supported becomes a problem in the future about
modeling metadata.

Hyrum's Law states:

    With a sufficient number of users of an API,
    it does not matter what you promise in the contract:
    all observable behaviors of your system
    will be depended on by somebody.

This is going to happen and we should nip this in the bud very quickly.
If metadata v2 should be allowed on charmhub it should be done in the
right way by following the spec, or not at all!


charmarchive.go
```diff
@@ -82,6 +83,20 @@ func readCharmArchive(zopen zipOpener) (archive *CharmArchive, err error) {
 		return nil, err
 	}
 
+	// If the format is not the v1 format (this should take care of any
+	// potential N formats), ensure that we can read the manifest file.
+	if b.meta.Format() != FormatV1 {
+		reader, err = zipOpenFile(zipr, "manifest.yaml")
+		if err != nil {
+			return nil, errors.Annotatef(err, "manifest file")
```
manadart, line 91: "opening manifest file"?

charmdir.go
```diff
@@ -66,45 +67,61 @@ func IsCharmDir(path string) bool {
...
 		return nil, errors.Annotatef(err, `issue parsing "metadata.yaml" file`)
 	}
 
-	file, err = os.Open(dir.join("config.yaml"))
+	// If the format is not the v1 format (this should take care of any
+	// potential N formats), ensure that we can read the manifest file.
+	if b.meta.Format() != FormatV1 {
+		reader, err = os.Open(b.join("manifest.yaml"))
+		if err != nil {
+			return nil, errors.Annotatef(err, `issue reading "manifest.yaml" file`)
```
manadart, line 89: Suggest dropping "issue ".

meta.go
```diff
@@ -810,124 +810,131 @@ func (c marshaledContainer) MarshalYAML() (interface{}, error) {
...
+	if m.Format() == FormatV1 {
+		for _, series := range m.Series {
+			if !IsValidSeries(series) {
+				return errors.Errorf("charm %q declares invalid series: %q", m.Name, series)
+			}
+		}
+	} else {
+		// Version 2 of the metadata should not delcare a series.
+		if len(m.Series) > 0 {
+			return errors.Errorf("charm %q declares both series and systems", m.Name)
```
manadart, line 884: I know upstream has not yet changed, but this will be "bases".

manadart (approved):
Looks good.

meta.go
SimonRichardson: Added a TODO.

SimonRichardson:
$$merge$$

PR #339, opened 2021-03-12
Block metadata v2 from being loaded in juju 2.8

hpidcock:
Backported version detection logic to prevent v2 metadata being loaded when juju 2.8 only supports v1 charms.

This is due to min-juju-version being removed for v2.


wallyworld (approved):

hpidcock:
$$merge$$

PR #340, opened 2021-03-12
Missing application body causes panic

SimonRichardson:
The following ensures that the base body is required when dealing with
overlays. The fix is really easy, but additionally I've added a test to
the bundle data to also check for empty bodies with ~ symbol.

Fixes: https://bugs.launchpad.net/juju/+bug/1918742

### DIFF
--- a/overlay.go
+++ b/overlay.go
@@ -440,6 +440,9 @@ func ReadAndMergeBundleData(sources ...BundleDataSource) (*BundleData, error) {
 		incResolver := sources[srcIndex].ResolveInclude
 		basePath := sources[srcIndex].BasePath()
 		for app, appData := range base.Data.Applications {
+			if appData == nil {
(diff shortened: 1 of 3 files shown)


overlay.go
```diff
@@ -440,6 +440,9 @@ func ReadAndMergeBundleData(sources ...BundleDataSource) (*BundleData, error) {
 		incResolver := sources[srcIndex].ResolveInclude
 		basePath := sources[srcIndex].BasePath()
 		for app, appData := range base.Data.Applications {
+			if appData == nil {
```
achilleasa, line 443: Should we also bail out if it's empty? Or will that interfere with the short-syntax work?

SimonRichardson: Bingo, I think I can use this for short syntax as well :)

achilleasa (approved):

SimonRichardson:
$$merge$$

PR #341, opened 2021-03-12
[v8] Missing application body causes panic

SimonRichardson:
The following ensures that the base body is required when dealing with
overlays. The fix is really easy, but additionally I've added a test to
the bundle data to also check for empty bodies with ~ symbol.

Fixes: https://bugs.launchpad.net/juju/+bug/1918742


achilleasa (approved):

SimonRichardson:
$$merge$$

PR #343, opened 2021-03-22
Bump juju/version dep

achilleasa:
This version bump drops the indirect dependency to gopkg.in/mgo.v2 in
favor to juju/mgo which this (and other) packages are using.


SimonRichardson (approved):
go mod static check would be nice.

achilleasa:
$$merge$$

PR #345, opened 2021-04-01
Encode base into a URL

SimonRichardson:
The following encodes a channel into a URL. A channel is comprised of
the following parts; track, risk, and branch. Track in this scenario is
always required and the other parts are optional.

To keep backwards compatibility with version 1 and HTTP version of the
charm URL, I have elected to keep the series. Series and channel are
exclusive and the String() method defines that. I believe we should
implement a validate for the URL to ensure that all components are
correct at the time of manual construction, but during parsing, it will
be impossible to have series and channel at the same time.

One last thing, I'm not sure about the split on the channel. As this is
intended to be a URL, I would like to keep it as close to a URL in terms
of semantics. So the URL should be parsable via url.Parse from the std
package, hence why we're not using #.

### DIFF
--- a/url.go
+++ b/url.go
@@ -81,15 +81,15 @@ type URL struct {
 	Name         string // "wordpress".
 	Revision     int    // -1 if unset, N otherwise.
 	Series       string // "precise" or "" if unset; "bundle" if it's a bundle.
-	Channel      string // "20.04:stable:branch" if channel is present, series is empty.
+	Base         string // "20.04:stable:branch" if base is present, series is empty.
 	Architecture string // "amd64" or "" if unset for charmstore (v1) URLs.
 }
 
 var (
-	validArch    = regexp.MustCompile("^[a-z]+([a-z0-9]+)?$")
-	validSeries  = regexp.MustCompile("^[a-z]+([a-z0-9]+)?$")
-	validChannel = regexp.MustCompile("^[a-zA-Z0-9.]+(:[a-zA-Z0-9]+(:[a-zA-Z0-9]+)?)?$")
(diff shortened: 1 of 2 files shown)


url.go
```diff
@@ -508,10 +535,12 @@ func parseHTTPURL(url *gourl.URL) (*URL, error) {
...
 //
 //  - ch:amd64/foo-1
-//  - ch:amd64/focal/foo-1
+//  - ch:amd64/20.04/foo-1
+//  - ch:amd64/20.04:stable/foo-1
+//  - ch:amd64/20.04:stable:branch/foo-1
 //  - ch:foo-1
 //  - ch:foo
-//  - ch:amd64/focal/foo
+//  - ch:amd64/20.04/foo
```
manadart, line 543: I don't think these examples are correct.

If you compare `ch:amd64/20.04:stable/foo-1` and `ch:amd64/windows:10/foo-1`, the middle section is _channel_ versus _name:channel_.

I think the struct should be:
```go
type URL struct {
	Schema       string // "cs", "ch" or "local".
	User         string // "joe".
	Name         string // "wordpress".
	Revision     int    // -1 if unset, N otherwise.
	Series       string // "precise" or "" if unset; "bundle" if it's a bundle.
	Base      string // "ubuntu:20.04:stable:branch" if base is present, series is empty.
	Architecture string // "amd64" or "" if unset for charmstore (v1) URLs.
}
```

Note base always has a name and track, with optional risk and branch.

SimonRichardson:
This is targetting the wrong base branch, opening a new PR.

PR #347, opened 2021-04-12
Add manifest details

SimonRichardson:
New BasesManifest method for CharmArchive and CharmDir, Manifest is
already used by CharmArchive.

Move Format and ComputedSeries from to CharmArchive and CharmDir,
both require manifest and metadata to compute.

Implementing ReadManifest and manifest Validate.

### DIFF
--- a/charm.go
+++ b/charm.go
@@ -17,10 +17,13 @@ var logger = loggo.GetLogger("juju.charm")
 // may be handled as a charm.
 type Charm interface {
 	Meta() *Meta
+	BasesManifest() *Manifest
 	Config() *Config
 	Metrics() *Metrics
 	Actions() *Actions
 	Revision() int
+	Format() Format
+	ComputedSeries() []string
 }
 
 // ReadCharm reads a Charm from path, which can point to either a charm archive or a
@@ -99,3 +102,13 @@ func IsUnsupportedSeriesError(err error) bool {
 	_, ok := err.(*unsupportedSeriesError)
 	return ok
 }
+
+// Format of the parsed charm.
+type Format int
+
+// Formats are the different versions of charm metadata supported.
+const (
+	FormatUnknown Format = iota
+	FormatV1      Format = iota
+	FormatV2      Format = iota
+)
--- a/charmarchive.go
+++ b/charmarchive.go
@@ -85,7 +85,7 @@ func readCharmArchive(zopen zipOpener) (archive *CharmArchive, err error) {
 
 	// If the format is not the v1 format (this should take care of any
 	// potential N formats), ensure that we can read the manifest file.
-	if b.meta.Format() != FormatV1 {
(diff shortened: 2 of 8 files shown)


charm.go
```diff
@@ -17,10 +17,13 @@ var logger = loggo.GetLogger("juju.charm")
 // may be handled as a charm.
 type Charm interface {
 	Meta() *Meta
+	BasesManifest() *Manifest
 	Config() *Config
 	Metrics() *Metrics
 	Actions() *Actions
 	Revision() int
+	Format() Format
```
SimonRichardson, line 25: Format shouldn't leak outside of `juju/charm`. Instead, we should ask the library what things it has based on the charm. An example of this would be `ComputedSeries`...

```go
func ComputedSeries(charm Charm) []string {
    ...
}
```

That would remove the need to constantly add methods to charm interface.

charm.go
```diff
@@ -99,3 +102,13 @@ func IsUnsupportedSeriesError(err error) bool {
 	_, ok := err.(*unsupportedSeriesError)
 	return ok
 }
+
+// Format of the parsed charm.
+type Format int
```
SimonRichardson, line 107: This should be an internal representation only.

charmarchive.go
```diff
@@ -385,3 +391,33 @@ func fixHookFunc(hooksDir string, hookNames map[string]bool) filepath.WalkFunc {
...
+func (a *CharmArchive) Format() Format {
+	if a.manifest.Bases != nil {
+		return FormatV2
+	}
+	return FormatV1
+}
+
+// ComputedSeries of a charm. This is to support legacy logic on new
+// charms that use Systems.
+func (a *CharmArchive) ComputedSeries() []string {
```
SimonRichardson, line 407: Make this a global function, based on the charm interface.

manifest.go
```diff
@@ -3,18 +3,118 @@
...
+	*m = Manifest{Bases: bases}
+	return nil
+}
+
+func parseBases(input interface{}) ([]systems.Base, error) {
+	var err error
+	if input == nil {
+		return nil, nil
+	}
+	res := []systems.Base(nil)
```
SimonRichardson, line 66:
```suggestion
	var res []systems.Base
```

manifest.go
```diff
@@ -3,18 +3,118 @@
...
+	for _, v := range input.([]interface{}) {
+		base := systems.Base{}
```
SimonRichardson, line 68:
```suggestion
		var base systems.Base
```

hmlanigan:
This was not ready for review, I've been reworking pieces.  

Format is one I'm unhappy with.  It was already leaking out from Meta.  Trying to determine a how to determine the format of metadata without having the manifest.  Or is it not possible.

hmlanigan:
replaced with https://github.com/juju/charm/pull/349

PR #349, opened 2021-04-13
Move bases from metadata to manifest

hmlanigan:
This commit is disruptive.   It introduces a new method, renames one, deletes one and moves one to a different type:

- CharmArchive Manifest is renamed to ArchiveMembers.
- The Charm interface added Manifest.
- Metadata ComputedSeries is moved to the CharmInterface.
- Metadata Format is removed.

ReadManifest has been implemented.  Bases now exist for Manifest, not Metadata.

Where Metadata is validated and check has split into 2 pieces, both in ReadMetadata and ReadCharm.  Full validation of Metadata now requires knowledge of a manifest if any exists.

Some compiler warning have been resolved as drive bys.

ToDo: use local channel and base structures instead of from the systems package.

### DIFF
--- a/charm.go
+++ b/charm.go
@@ -18,6 +18,7 @@ var logger = loggo.GetLogger("juju.charm")
 type Charm interface {
 	Meta() *Meta
 	Config() *Config
+	Manifest() *Manifest
 	Metrics() *Metrics
 	Actions() *Actions
 	Revision() int
--- a/charm.go
+++ b/charm.go
@@ -22,6 +22,7 @@ type Charm interface {
 	Metrics() *Metrics
 	Actions() *Actions
 	Revision() int
+	ComputedSeries() []string
 }
 
 // ReadCharm reads a Charm from path, which can point to either a charm archive or a
--- a/charm.go
+++ b/charm.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/juju/loggo"
+	"github.com/juju/collections/set"
 )
 
 var logger = loggo.GetLogger("juju.charm")
@@ -22,7 +23,6 @@ type Charm interface {
 	Metrics() *Metrics
 	Actions() *Actions
 	Revision() int
-	ComputedSeries() []string
 }
 
 // ReadCharm reads a Charm from path, which can point to either a charm archive or a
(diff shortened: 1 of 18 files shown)


charm.go
```diff
@@ -38,7 +40,15 @@ func ReadCharm(path string) (charm Charm, err error) {
 	if err != nil {
 		return nil, err
 	}
-	return charm, nil
+
+	// Find out the charm format, to Check the metadata.  It should
+	// be one format or the other.
+	format := FormatV2
```
manadart, line 46: What is the feasibility of `Charm.Meta()` returning an interface instead of a reference?

If we did it this way we could:
- Have `MetaV1` and `MetaV2` as implementations, all embedding a `MetaBase`, but having different `Check`.
- Determine the format and therefore the implementation in `ReadCharm`
- Remove `checkV1` and `checkV2`.

charmarchive.go
```diff
@@ -251,6 +252,32 @@ func (a *CharmArchive) LXDProfile() *LXDProfile {
...
 }
 
+// Manifest returns the Manifest representing the manifest.yaml file
+// for the charm expanded in dir.
+func (a *CharmArchive) Manifest() *Manifest {
+	return a.manifest
+}
+
+// ComputedSeries of a charm. This is to support legacy logic on new
+// charms that use Systems.
```
manadart, line 262: Systems or Bases?

charmdir.go
```diff
@@ -246,6 +248,32 @@ func (dir *CharmDir) LXDProfile() *LXDProfile {
 	return dir.lxdProfile
 }
 
+// Manifest returns the Manifest representing the manifest.yaml file
+// for the charm expanded in dir.
+func (dir *CharmDir) Manifest() *Manifest {
```
manadart, line 253: Not sure why we don't embed the common elements and let methods like these get hoisted.

charmdir.go
```diff
@@ -246,6 +248,32 @@ func (dir *CharmDir) LXDProfile() *LXDProfile {
...
+func (dir *CharmDir) ComputedSeries() []string {
+	if len(dir.manifest.Bases) == 0 {
+		return dir.meta.Series
+	}
+	// The slice must be ordered based on system appearance but
+	// have unique elements.
+	seriesSlice := []string(nil)
+	seriesSet := set.NewStrings()
+	for _, base := range dir.manifest.Bases {
+		series := base.String()
```
manadart, line 268: I don't like the fact that a stringified base is a series. I will look at that when I bring base in from systems.

manadart (approved):
Some suggestions, but this is good for advancing us.

meta.go
```diff
@@ -1628,15 +1410,10 @@ func identifyFormat(raw map[interface{}]interface{}) (Format, error) {
...
 		}
 	}
-	if format == FormatUnknown {
-		//return FormatUnknown, errors.Errorf("unknown metadata format")
-		// TODO: return FormatUnknown when unit tests are upgraded to both v1 and v1 metadata.
-		return FormatV1, nil
-	}
 	if mismatched != nil {
-		return FormatUnknown, errors.Errorf("ambigious metadata: keys %s cannot be used with %s",
+		return errors.Errorf("ambigious metadata: keys %s cannot be used with %s",
```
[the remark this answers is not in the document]
hmlanigan: done

charmarchive.go
hmlanigan: done

charmdir.go
hmlanigan: done

hmlanigan:
$$merge$$

charmdir.go
```diff
@@ -45,14 +45,7 @@ var defaultJujuIgnore = `
...
 	Path       string
-	meta       *Meta
-	config     *Config
-	metrics    *Metrics
-	actions    *Actions
-	lxdProfile *LXDProfile
-	manifest   *Manifest
-	revision   int
-	version    string
+	*charmBase
```
manadart, line 48: :heart:

PR #351, opened 2021-04-14
Add base structure

hmlanigan:
Imports base logic from juju/systems to this package.  Will be consumed by this package and juju in the future.  There are a
few important changes.

A base now knows nothing about a series, and does not try to interpret or provide one base on given data.  The base string does not return an interpretation of the series from the base, it is simply a string representation of the base with slash delimiter.

The name is changed to lower case where possible.  Issues have been seen in juju making requests to the Charmhub API.

Use os pkg version of operating systems instead of copying over.

Add an Architectures slice to Base for the spec.  Be nice with naming and use juju utils arch pkg helpers.


base.go
```diff
@@ -0,0 +1,121 @@
...
+	err = base.Validate()
+	if err != nil {
+		return Base{}, errors.Annotatef(err, "invalid base string %q", s)
+	}
+	return base, nil
+}
+
+// ParseBaseWithArchitectures parses a base as string in the form
+// "os/track/risk/branch" and a list of architectures
+func ParseBaseWithArchitectures(s string, archs []string) (Base, error) {
```
hmlanigan, line 73: We may want just one ParseBase, including architectures in the single string would be unwieldy.

channel.go
```diff
@@ -54,9 +54,9 @@ func isRisk(potential string) bool {
 //    <track>/<risk>/<branch>
 //
 type Channel struct {
-	Track  string
-	Risk   Risk
-	Branch string
+	Track  string `json:"track,omitempty"`
```
hmlanigan, line 57: Needed for TestJSONEnconding

charm.go
```diff
@@ -76,7 +76,8 @@ func SeriesForCharm(requestedSeries string, supportedSeries []string) (string, e
 }
 
 // ComputedSeries of a charm. This is to support legacy logic on new
-// charms that use Bases.
+// charms that use Bases.  For v1 metadata, series are returned. For
+// v2 metadata, the tracks of channels in bases are returned.
```
hmlanigan, line 80: We'll need to be careful with this when consumed by juju

base.go
manadart: I think one method with a variadic arg for architectures would work.

hmlanigan: thanks

hmlanigan:
!!build!!

hmlanigan:
!!build!!

charm.go
```diff
@@ -93,7 +94,7 @@ func ComputedSeries(c CharmMeta) []string {
 	seriesSlice := []string(nil)
 	seriesSet := set.NewStrings()
 	for _, base := range manifest.Bases {
-		series := base.String()
+		series := base.Channel.Track
```
manadart, line 97: This will break all the places it is used in Juju. We need utilise a method that gets a series based on the base.

manadart (changes requested):
I think we need to preserve the old `ComputedSeries` behaviour.

charm.go
hmlanigan: Let's chat.  I do not like having series conversion in the charm package.  We've been pulling it into juju. There are also issues with finding new versions etc, we've worked hard in juju to avoid.  Looks like a backwards step.  

We do need a function to convert in juju.

manadart: As discussed, `ComputedSeries` moves to Juju.

manadart (approved):
Good to move forward.

hmlanigan:
!!build!!

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #353, opened 2021-04-15
Ensure we don't panic on a nil manifest

SimonRichardson:
Given older charms that already live in state, it's possible that you
can end up with no manifest at all. If that's the case we need to ensure
that when you compute a series it doesn't panic.


manadart (approved):

SimonRichardson:
$$merge$$

PR #355, opened 2021-04-15
Encapsulate version determination and metadata check

manadart:
Adds a method for checking charm metadata in a single step, including the version determination.
This allows a single method call in Juju at all the sites where we currently call `Meta.Check()`.

### DIFF
--- a/charm.go
+++ b/charm.go
@@ -47,14 +47,20 @@ func ReadCharm(path string) (charm Charm, err error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


manadart:
[charm version/check encapsulation](https://trello.com/c/GJMiN8az/3613-charm-version-check-encapsulation)

hmlanigan (approved):

manadart:
$$merge$$

PR #357, opened 2021-04-16
bson omitempty for Base elements.

hmlanigan:
add bson omitempty to elements of bases, for proper handling in the juju db.


jameinel (approved):
This change is correct, but it symptomatic of why we shouldn't be serializing charm objects to the database directly.

jameinel (approved):

hmlanigan:
$$merge$$

PR #360, opened 2021-04-22
Expose validate error

SimonRichardson:
The error during validation was being trapped, we need to ensure that
the error bubbles up so that we can show the issue to the user.

### DIFF
--- a/manifest.go
+++ b/manifest.go
@@ -29,7 +29,7 @@ func NewManifest() *Manifest {
 func (m *Manifest) Validate() error {
 	for _, b := range m.Bases {
 		if err := b.Validate(); err != nil {
-			return errors.Annotate(err, "invalid base")
+			return errors.Annotate(err, "validating manifest")
 		}
 	}
 	return nil
--- a/manifest.go
+++ b/manifest.go
@@ -4,7 +4,6 @@
 package charm
 
 import (
-	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


manifest.go
```diff
@@ -30,7 +29,7 @@ func NewManifest() *Manifest {
 func (m *Manifest) Validate() error {
 	for _, b := range m.Bases {
 		if err := b.Validate(); err != nil {
-			return fmt.Errorf("invalid base: empty file")
+			return errors.Annotate(err, "invalid base")
```
manadart, line 32: "validating base"

manifest_test.go
```diff
@@ -44,3 +44,21 @@ bases:
...
 	}})
 }
+
+func (s *manifestSuite) TestReadValidateManifest(c *gc.C) {
+	_, err := ReadManifest(strings.NewReader(`
+bases:
+  - name: ""
+    channel: "18.04"
+`))
+	c.Assert(err, gc.ErrorMatches, "manifest: name must be specified not valid")
```
manadart, line 54: Ultimate emission should be "base without name not valid".

manadart (changes requested):
Wording suggestions.

manadart (approved):
Much better.

SimonRichardson:
$$merge$$

PR #361, opened 2021-04-22
Expose validate error

SimonRichardson:
The error during validation was being trapped, we need to ensure that
the error bubbles up so that we can show the issue to the user.


manadart (approved):

SimonRichardson:
$$merge$$

PR #365, opened 2021-05-04
Change EnsureSchema signature to accept the default schema as an arg

achilleasa:
The default schema that should be used depends on the controller that
juju is talking to. Therefore, it should not be specified as a
package-level public variable and instead should be explicitly provided
by the caller when calling EnsureSchema.

### DIFF
--- a/url.go
+++ b/url.go
@@ -50,12 +50,6 @@ func (s Schema) String() string {
 	return string(s)
 }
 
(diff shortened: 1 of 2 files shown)


hmlanigan (approved):
LGTM

achilleasa:
$$merge$$

PR #366, opened 2021-05-10
Resolve out of sync presence map

SimonRichardson:
The following attempts to keep the presence map in sync with the actual
data after normalization. During a normalization "scale" is swapped
out for "num_units". This is fine, except when it comes to overlays you
now don't know if scale or num_units was used for the overlays and
subsequent scale calls are now not used.

This is actually a problem with the library itself. There should be an
internal representation of a bundle that carries all this important
information, which should include schemas etc. At the very end of the
reading/parsing process, we should hand back a clean version of that
charm/bundle. That way we can make the bundle grow organically without
the requirement of so many hacks.

The code is simple, detect and resolve a normalization and just insert
the right value into the presence map before applying an overlay.

-----

Bug: https://bugs.launchpad.net/juju/+bug/1884465

### DIFF
--- a/overlay.go
+++ b/overlay.go
@@ -512,8 +512,8 @@ func resolveOverlayPresenceFields(base *BundleDataPart) {
 		}
 
 		presence := applications.forField(name)
-		// If the presence map contains scale, but doesn't containt num_units
-		// and if the app.Scale_ has be set to zero. We can then assume that a
+		// If the presence map contains scale, but doesn't contain num_units
+		// and if the app.Scale_ has been set to zero. We can then assume that a
 		// normalistion has occurred.
 		if presence.fieldPresent("scale") && !presence.fieldPresent("num_units") && app.Scale_ == 0 && app.NumUnits > 0 {
 			presence["num_units"] = presence["scale"]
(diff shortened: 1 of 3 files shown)


overlay.go
```diff
@@ -490,6 +494,33 @@ func ReadAndMergeBundleData(sources ...BundleDataSource) (*BundleData, error) {
...
+	}
+	for name, app := range base.Data.Applications {
+		if !applications.fieldPresent(name) {
+			continue
+		}
+
+		presence := applications.forField(name)
+		// If the presence map contains scale, but doesn't containt num_units
+		// and if the app.Scale_ has be set to zero. We can then assume that a
+		// normalistion has occurred.
```
achilleasa, line 517:
```suggestion
		// If the presence map contains scale, but doesn't contain num_units
		// and if the app.Scale_ has been set to zero. We can then assume that a
		// normalisation has occurred.
```

Also `normalization` vs `normalisation`?

overlay.go
```diff
@@ -490,6 +494,33 @@ func ReadAndMergeBundleData(sources ...BundleDataSource) (*BundleData, error) {
...
+		if presence.fieldPresent("scale") && !presence.fieldPresent("num_units") && app.Scale_ == 0 && app.NumUnits > 0 {
```
achilleasa, line 518: I wonder whether there is a check somewhere `if scale > 0 && num_units >0 { // error }`. You shouldn't be specifying both at the same time right?

achilleasa: Oh, just saw your test below. I think we should add the check as a drive-by. Otherwise, there is an ambiguity in the actual value for num units that must be resolved by the user.

achilleasa (approved):
LGTM (needs some tweaking with the comment typos). However, I think it would be a good idea to add a check as suggested in the PR comments.

SimonRichardson:
$$merge$$

PR #372, opened 2021-06-11
Do not set an application series to kubernetes if not specified.

hmlanigan:
Let juju determine what the series should be for kubernetes charms in a bundle.  The default is no longer kubernetes with the addition of sidecar charms.  See LP1928796.  Causes issues with local sidecar charms.


SimonRichardson (approved):

SimonRichardson:
@hmlanigan quickly, I remembered this now, how do we not run into this https://github.com/juju/charm/blob/v8/bundledata.go#L664

hmlanigan:
$$merge$$

PR #374, opened 2021-07-16
Format selection reasons

SimonRichardson:
The following starts to implement the reasons why a format was selected.
We can use that information to improve the error messages for users
to better understand.

Work towards fix the following bug: https://bugs.launchpad.net/juju/+bug/1936281


charmarchive.go
```diff
@@ -81,7 +81,7 @@ func readCharmArchive(zopen zipOpener) (archive *CharmArchive, err error) {
 	// this charm is v1 or not.
 	reader, err = zipOpenFile(zipr, "manifest.yaml")
 	if _, ok := err.(*noCharmArchiveFile); ok {
-		b.manifest = NewManifest()
```
manadart, line 84: Do we still need `NewManifest` if it's never non-empty?

manadart (approved):
Looks good.

charmarchive.go
SimonRichardson: No idea, will check.

SimonRichardson:
$$merge$$

PR #377, opened 2021-07-22
Charm+format: Series always dictates a format v1

SimonRichardson:
If a charm contains a series, even with the presence of manifest.yaml
then that charm is a v1. This is a damn shame, because our modelling
becomes weaker inside of juju, but it seems like charmcraft doesn't
restrict uploads with series and manifests.

The fix is a bit more involved, as we can use a set.String for better
look up times. It also removes duplications with in the selection
reasons.

### DIFF
--- a/charm.go
+++ b/charm.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/juju/collections/set"
 	"github.com/juju/errors"
 	"github.com/juju/loggo"
 )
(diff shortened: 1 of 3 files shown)


charm.go
```diff
@@ -63,31 +64,37 @@ const (
 	SelectionContainers FormatSelectionReason = "containers"
 )
 
+var (
```
manadart, line 67: Really need the braces here?

manadart (approved):

charm.go
SimonRichardson: I was going to add a bigger comment but didn't in the end.

SimonRichardson:
$$merge$$

PR #382, opened 2021-10-05
Parse assumes syntax

achilleasa:
This PR implements the proposed syntax for `assumes` blocks as described in the [spec](https://docs.google.com/document/d/1D4ECIWFBxe_7IZE0q1nkEGE8O2QqmTinKOpx5EL7ReI/edit#).

Due to the complex nature of the syntax, the PR provides `assume.TreeExpression`, a wrapper with custom YAML marshal/unmarshal logic to aid with the parsing and validation of the assumes block.


assumes/parser.go
```diff
@@ -0,0 +1,257 @@
...
+	var numPeriods int
+	for _, ch := range verStr {
+		if ch == '.' {
+			numPeriods++
+		}
+	}
+
+	for i := numPeriods; i < 2; i++ {
+		verStr += ".0"
+	}
```
achilleasa, line 185: Nasty hack ⚠️

assumes/expression.go
```diff
@@ -0,0 +1,61 @@
...
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package assumes
+
+import "github.com/juju/version/v2"
+
+var (
+	_ Expression = (*FeatureExpression)(nil)
+	_ Expression = (*CompositeExpression)(nil)
+)
```
SimonRichardson, line 11: I never get why these aren't in test files? As they're asserts on the implementations, you don't need them to be checked every time you run.

assumes/expression.go
```diff
@@ -0,0 +1,61 @@
...
+type Expression interface {
+	Type() ExpressionType
+}
+
+// VersionConstraint describes a constraint for required feature versions.
+type VersionConstraint string
+
+const (
+	VersionGTE VersionConstraint = ">="
+	VersionLT  VersionConstraint = "<"
```
SimonRichardson, line 31: I understand why you don't want `=`, as that exposes locked files, but why no LTE, that would seem useful?

assumes/parser.go
```diff
@@ -0,0 +1,257 @@
...
+			rawVersion: versionStr,
+		}, nil
+	}
+
+	return nil, errors.Errorf("malformed feature expression %q", exprString)
+}
+
+func parseVersion(verStr string) (*version.Number, error) {
+	// Try to parse the version string as-is
+	ver, pErr := version.Parse(verStr)
```
SimonRichardson, line 167: I really think we should just fix this. `version.Parse` claims to do something, which it clearly doesn't.

SimonRichardson (commented):
I still think we should rev version package to v3 and make it do what we think it should do!

assumes/expression.go
achilleasa: These ones are compile-time assertions; they don't trigger allocations.

assumes/expression.go
achilleasa: This was debated in the spec review session. Typically, you know the bare minimum version (hence the `>=`) and you now the version where it _stops_ working (hence the `<`) here.

assumes/expression.go
```diff
@@ -0,0 +1,61 @@
...
+	// The name of the feature.
+	Name string
+
+	// A feature within an assumes block may optionally specify a version
+	// constraint.
+	Constraint VersionConstraint
+	Version    *version.Number
+
+	// The raw, unprocessed version string for serialization purposes.
+	rawVersion string
```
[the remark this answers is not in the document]
achilleasa: `version.Number` will expand the version string to a qualified semver value (with `0` for the missing parts). When serializing back we want to preserve the original version as specified by the charm author so we need to record that.

For example, if the assumes expression specifies `1.8` we don't want it to become `1.8.0` because that is not necessarily the same as the version originally specified.

assumes/parser.go
achilleasa: While the regex claims it can be used to parse different formats (which is indeed incorrect), the [docblock](https://github.com/juju/version/blob/master/version.go#L162-L164) on `version.Parse` mentions that it assumes `major.minor.patch` format.

As we use that package all over the place I am a bit hesitant to patch that particular method.

assumes/expression.go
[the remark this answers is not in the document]
SimonRichardson: Then we shouldn't be using version.Number.

assumes/expression.go
achilleasa: Also you don't want them in tests because you could still build a binary without running the tests

assumes/parser.go
achilleasa: Hack-b-gone!

PR #401, opened 2022-08-12
[JUJU-1606] 1984061 default channel vs latest

hmlanigan:
Stop removing "latest" from a normalized channel.

Related to fix for https://bugs.launchpad.net/juju/+bug/1984061


juanmanuel-tirado (approved):
LG2M

v8.0.3 should be released after merging.

PR #402, opened 2022-08-29
[JUJU-1584] Adjust wording of user friendly yaml unmarshall errors even further

hmlanigan:
To address comments in [juju pr 14474](https://github.com/juju/juju/pull/14474): Adjust wording of user friendly yaml unmarshall errors even further. Result can be seen in the unit test.

### DIFF
--- a/bundledatasrc.go
+++ b/bundledatasrc.go
@@ -5,10 +5,12 @@ package charm
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/juju/errors"
@@ -230,8 +232,7 @@ func parseBundleParts(r io.Reader) ([]*BundleDataPart, error) {
 			break
 		} else if err != nil {
 			if strings.HasPrefix(err.Error(), "yaml: unmarshal errors:") {
-				friendlyErrors := userFriendlyUnmarshalErrors(err)
-				part.UnmarshallError = errors.Annotatef(friendlyErrors, "unmarshal document %d", docIdx)
+				part.UnmarshallError = userFriendlyUnmarshalErrors(errors.Annotatef(err, "document %d", docIdx))
 			} else {
 				return nil, errors.Annotatef(err, "unmarshal document %d", docIdx)
 			}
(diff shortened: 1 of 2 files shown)


bundledatasrc.go
```diff
@@ -245,13 +247,18 @@ func parseBundleParts(r io.Reader) ([]*BundleDataPart, error) {
...
+// userFriendlyUnmarshalErrors provides user friendly versions of the
+// yaml unmarshall errors for a bundle.
+// Lines such as: "line 1: field name not found in type charm.ApplicationSpec"
+// are converted to: "line 1: unrecognized field name"
+// The actual errors for developers can be found with Trace level logging
+// enabled.
 func userFriendlyUnmarshalErrors(err error) error {
 	logger.Tracef("developer friendly error message: \n%s", err.Error())
 	friendlyText := err.Error()
-	friendlyText = strings.ReplaceAll(friendlyText, "type charm.ApplicationSpec", "applications")
```
jameinel, line 251: Just to make sure I understand this change:
Is this where the 'not found in applications' came from? eg it originally said 'unmarshal errors: field foo not found in type charm.ApplicationSpec'.
I think what you've done is definitely better. I wonder if we would rather: `" unrecognized field: "` because that gives us a slightly clearer: 
```
  line 5: unrecognized field: constrain
```
I would normally just quote it with:
```
  line 5: unrecognized field "constrain"
```

But I feel like some sort of character to distinguish what part of the sentence is the field name is useful.

jameinel (approved):
Thanks for this. I definitely think the errors read a lot better now. One small tweak for a ':' or  quoting to clearly indicate the field name, but otherwise great.

hmlanigan:
squashed the commits to 1

bundledatasrc_test.go
```diff
@@ -133,14 +133,14 @@ applications:
...
+		"document 0:\n"+
+		"  line 5: unrecognized field \"constrain\"\n"+
+		"  line 8: unrecognized field \"num_uns\"")
 	c.Assert(parts[1].UnmarshallError, jc.ErrorIsNil)
 	c.Assert(parts[2].UnmarshallError, gc.NotNil)
 	c.Assert(parts[2].UnmarshallError.Error(), gc.Matches, ""+
-		"unmarshal document 2: yaml: unmarshal errors:\n"+
-		"  line 21: field offer not found in applications")
+		"document 2:\n"+
+		"  line 21: unrecognized field \"offer\"")
```
jameinel, line 143: As an aside, it is often nice to use `` as a way to get around the extra quoting, etc. eg:
```
	c.Assert(parts[2].UnmarshallError.Error(), gc.Matches, `
document 2:
  line 21: unrecognized field "offer"
`[1:])
```

(The [1:] magic is because you want to start left aligned but that means an opening '\n' character)

jameinel (approved):
I don't know if you'd find the other string formatting cleaner, but I like how these error messages look.
Thanks for the effort.

PR #405, opened 2022-09-08
[JUJU-1691] Url parse first part can be series

cderici:
#### Description

This change fixes the bug in the `charm.ParseURL`, where the first part of an identifier URL is assumed to be the architecture, whereas it can also be the series. As a result:

`ch:focal/istio-gateway-74` is parsed as `{Schema:"ch" User: "" Name:"istio-gateway" Revision:74 Series: "" Architecture:"focal"}`

#### QA Steps

I added the case above in the ParseURL unit tests, along with the `ch:amd64/istio-gateway-74` case. You may just run that:

```
go test -gocheck.v -gocheck.f TestParseURL
```

### DIFF
--- a/url.go
+++ b/url.go
@@ -118,8 +118,8 @@ func ValidateSeries(series string) error {
 
 // IsValidArchitecture reports whether the architecture is a valid architecture
 // in charm or bundle URLs.
-func IsValidArchitecture(arch string) bool {
-	return validArch.MatchString(arch)
+func IsValidArchitecture(architecture string) bool {
+	return validArch.MatchString(architecture) && arch.IsSupportedArch(architecture)
 }
 
 // ValidateArchitecture returns an error if the given architecture is invalid.
--- a/url.go
+++ b/url.go
@@ -196,7 +196,6 @@ func MustParseURL(url string) *URL {
 func ParseURL(url string) (*URL, error) {
 	// Check if we're dealing with a v1 or v2 URL.
 	u, err := gourl.Parse(url)
-
 	if err != nil {
 		return nil, errors.Errorf("cannot parse charm or bundle URL: %q", url)
 	}
--- a/url.go
+++ b/url.go
@@ -14,6 +14,7 @@ import (
 	"github.com/juju/errors"
 	"github.com/juju/mgo/v3/bson"
 	"github.com/juju/names/v4"
+	"github.com/juju/utils/v3/arch"
 )
 
 // Schema represents the different types of valid schemas.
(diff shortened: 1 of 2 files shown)


url.go
```diff
@@ -542,12 +554,12 @@ func parseIdentifierURL(url *gourl.URL) (*URL, error) {
...
 	if r.Architecture != "" {
 		if err := ValidateArchitecture(r.Architecture); err != nil {
-			return nil, errors.Annotatef(err, "cannot parse architecture in URL %q", url)
+			return nil, errors.Annotatef(err, "cannot parse architecture/series in URL %q", url)
 		}
 	}
 	if r.Series != "" {
 		if err := ValidateSeries(r.Series); err != nil {
-			return nil, errors.Annotatef(err, "cannot parse series in URL %q", url)
+			return nil, errors.Annotatef(err, "cannot parse architecture/series in URL %q", url)
```
hmlanigan, line 562: After reading the full error below in a test:
```suggestion
			return nil, errors.Annotatef(err, "in URL %q", url)
```

url.go
```diff
@@ -542,12 +554,12 @@ func parseIdentifierURL(url *gourl.URL) (*URL, error) {
 	// Optional
 	if r.Architecture != "" {
 		if err := ValidateArchitecture(r.Architecture); err != nil {
```
hmlanigan, line 556: ValidateArchitecture doesn't check IsSupportedArch, it only does a regex for strings that start with an alphabetic character and end with an alpha/numeric character. We should check IsSupported in both cases, where there is only an arch or both a series and arch, to be consistent.

url.go
```diff
@@ -542,12 +554,12 @@ func parseIdentifierURL(url *gourl.URL) (*URL, error) {
...
-			return nil, errors.Annotatef(err, "cannot parse architecture in URL %q", url)
+			return nil, errors.Annotatef(err, "cannot parse architecture/series in URL %q", url)
```
hmlanigan, line 557: After reading the full error below in a test:
```suggestion
			return nil, errors.Annotatef(err, "in URL %q", url)
```

hmlanigan (changes requested):

url.go
```diff
@@ -195,6 +196,7 @@ func MustParseURL(url string) *URL {
 func ParseURL(url string) (*URL, error) {
 	// Check if we're dealing with a v1 or v2 URL.
 	u, err := gourl.Parse(url)
+
```
hmlanigan, line 199: rm new line

hmlanigan (commented):
LGTM

hmlanigan (approved):
Clicking on approved this time. :-)

PR #407, opened 2022-10-19
[JUJU-2002] Fix LP1984061. "latest" track is not the default track.

hmlanigan:
Stop removing "latest" from a normalized channel.

Port of #401 into master, v9


cderici (approved):

jameinel:
Thanks. I'm glad this has landed.

PR #408, opened 2023-01-03
[JUJU-2285] Update package to revision 10 and remove charmstore as a valid URL schema

hmlanigan:
Remove "cs" has a valid schema for a charm url. Part of removing charm store support from juju. Includes removal of "http" and "https" as they are resolved to a "cs".

Given this is a breaking change, update the package revision to 10.

All unit tests should pass, including the new one testing "cs" is not a valid schema.


url.go
```diff
@@ -210,10 +187,6 @@ func ParseURL(url string) (*URL, error) {
 	case u.Opaque != "":
 		u.Path = u.Opaque
 		curl, err = parseV1URL(u, url)
```
SimonRichardson, line 189: Do local charm URLs hit this, I'm not sure what uses this anymore?

url.go
```diff
@@ -573,7 +465,7 @@ func EnsureSchema(url string, defaultSchema Schema) (string, error) {
 		return "", errors.Errorf("cannot parse charm or bundle URL: %q", url)
 	}
 	switch Schema(u.Scheme) {
-	case CharmStore, CharmHub, Local, HTTP, HTTPS:
+	case CharmHub, Local, HTTP, HTTPS:
```
SimonRichardson, line 468: Remove HTTP, HTTPS.

url.go
```diff
@@ -21,13 +21,10 @@ import (
...
-	CharmStore Schema = "cs"
-
 	// Local represents a local charm URL, describes as a file system path.
 	Local Schema = "local"
 
-	// CharmHub schema represents the new schema for another unique charm store.
+	// CharmHub schema represents the charmhub charm repository.
 	CharmHub Schema = "ch"
 
 	// HTTP refers to the HTTP schema that is used for the V2 of the charm URL.
```
SimonRichardson, line 30: Remove HTTP and HTTPS. It never escapes into juju itself (HTTP(S) is converted into cs URLS), so it makes sense to nuke it here as well.

url.go
```diff
@@ -60,16 +57,10 @@ type Location interface {
...
-//     cs:development/precise/wordpress-20
-//     cs:~joe/development/wordpress
-//     ch:wordpress
-//
+//	local:oneiric/wordpress
+//	cs:wordpress
+//	ch:wordpress
+//	ch:amd64/jammy/wordpress-30
 type URL struct {
 	Schema       string // "cs", "ch" or "local".
```
SimonRichardson, line 65: Remove `cs`.

url.go
```diff
@@ -88,15 +79,14 @@ var (
 // ValidateSchema returns an error if the schema is invalid.
 //
 // Valid schemas for the URL are:
-// - cs: charm store
 // - ch: charm hub
 // - local: local file
 //
 // http and https are not valid schemas, as they compiled to V1 charm URLs.
```
SimonRichardson, line 85: Remove http and https comment.

url.go
```diff
@@ -229,9 +202,7 @@ func ParseURL(url string) (*URL, error) {
 }
 
 func parseV1URL(url *gourl.URL, originalURL string) (*URL, error) {
-	r := URL{
-		Schema: CharmStore.String(),
-	}
+	r := URL{}
 	if url.Scheme != "" {
 		r.Schema = url.Scheme
 	}
```
SimonRichardson, line 208: Officially, we can now close the weird hole here. Previously, if you passed in `ftp://` it would put that as the front of the charm URL. I don't think we'll ever want that. Instead, I think we should be a lot stricter and error out when we don't have a local scheme. This was originally how they squeezed in `http` and `https`, which I was never fond of. 

```go

	if url.Scheme != "local" {
		return nil, errors.Errorf("expected local scheme: %q", url.Scheme)
	}
    r := URL{}
```

SimonRichardson (changes requested):
Let's nuke http and https completely.

url.go
hmlanigan: Thank you.

SimonRichardson (approved):

PR #410, opened 2023-05-17
[JUJU-3479] Support bases in bundles

jack-w-shaw:
We migrating away from using series, towards bases instead. Add support for specifying bases for applications, machines, and the bundle default

Also, as a flyby, fix bases to use the new format with OS and channel separated with an at sign instead of a slash

### QA steps

Verify unit tests pass

Compile into juju after applying the following diff:
```
diff --git a/api/common/charms/common.go b/api/common/charms/common.go
index d0f726300b..c8a2982229 100644
--- a/api/common/charms/common.go
+++ b/api/common/charms/common.go
@@ -394,7 +394,7 @@ func convertCharmManifest(input *params.CharmManifest) (*charm.Manifest, error)
        }
        res := []charm.Base(nil)
        for _, v := range input.Bases {
-               str := fmt.Sprintf("%s/%s", v.Name, v.Channel)
+               str := fmt.Sprintf("%s@%s", v.Name, v.Channel)
                b, err := charm.ParseBase(str, v.Architectures...)
                if err != nil {
                        return nil, errors.Trace(err)
```

And attempt to deploy the following bundles:
```
$ cat bundle.yaml
name: ubu
applications:
  ubuntu:
    charm: ubuntu
    scale: 1
    base: ubuntu@22.04
$ juju deploy ./bundle.yaml
```
Should deploy successfully without error. NOTE base has not yet been implemented on the Juju side. Do not pay attention to the base the application is actually deployed to

```
$ cat bundle.yaml
name: ubu
applications:
  ubuntu:
    charm: ubuntu
    scale: 1
    base: ubuntu
$ juju deploy ./bundle.yaml
ERROR cannot deploy bundle: the provided bundle has the following errors:
application "ubuntu" declares an invalid base "ubuntu"
```

### DIFF
--- a/bundledata.go
+++ b/bundledata.go
@@ -47,8 +47,15 @@ type BundleData struct {
 	Saas map[string]*SaasSpec `bson:"saas,omitempty" json:"saas,omitempty" yaml:"saas,omitempty"`
 
 	// Series holds the default series to use when
-	// the bundle chooses charms.
+	// the bundle deploys applications. A series defined for an application
+	// takes precedence.
+	// Series and Base cannot be mixed.
 	Series string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+
+	// Base holds the default base to use when the bundle deploys
+	// applications. A base defined for an application takes precedence.
+	// Series and Base cannot be mixed.
+	DefaultBase string `bson:"default-base,omitempty" json:"default-base,omitempty" yaml:"default-base,omitempty"`
 
 	// Relations holds a slice of 2-element slices,
 	// each specifying a relation between two applications.
@@ -79,6 +86,7 @@ type MachineSpec struct {
 	Constraints string            `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 	Annotations map[string]string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 	Series      string            `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
+	Base        string            `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 }
 
 // ApplicationSpec represents a single application that will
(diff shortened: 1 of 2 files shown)


bundledata.go
```diff
@@ -50,6 +50,10 @@ type BundleData struct {
 	// the bundle chooses charms.
 	Series string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 
+	// Base holds the default base to use when the bundle chooses
+	// charms. Base should overwrite Series
```
hmlanigan, line 54: We should error if both are provided. Otherwise the user may have unexpected results.

jack-w-shaw: Resolved in latest commit

jack-w-shaw:
@SimonRichardson I have dropped that change from this PR

So only a patch rev is required

bundledata.go
```diff
@@ -48,8 +48,14 @@ type BundleData struct {
 
 	// Series holds the default series to use when
 	// the bundle chooses charms.
+	// Series and Base cannot be mixed in a bundle
 	Series string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
 
+	// Base holds the default base to use when the bundle chooses
+	// charms.
+	// Series and Base cannot be mixed in a bundle
+	Base string `bson:",omitempty" json:",omitempty" yaml:",omitempty"`
```
hmlanigan, line 57: We talked about making this "Default-Base" during our meeting for clarity.

hmlanigan (commented):
Wanted to get comments out. Haven't QA'd

bundledata.go
```diff
@@ -535,6 +543,12 @@ func (bd *BundleData) verifyBundle(
 	if bd.Series != "" && !IsValidSeries(bd.Series) {
 		verifier.addErrorf("bundle declares an invalid series %q", bd.Series)
 	}
+	if bd.Base != "" {
+		if _, err := ParseBase(bd.Base); err != nil {
```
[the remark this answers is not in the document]
jack-w-shaw: Are there? Not in juju/charm that I can find

PR #411, opened 2023-05-18
[JUJU-3749] Update base format

jack-w-shaw:
Update base format to support bases which look like `ubuntu@22.04/stable`

Previously we used bases like `ubuntu/22.04/stable`

NOTE: This is a breaking change, so we will need a major version rev

### QA Steps

Ensure unit tests pass

Compile into juju after applying the following diff:
```
diff --git a/api/common/charms/common.go b/api/common/charms/common.go
index d0f726300b..c8a2982229 100644
--- a/api/common/charms/common.go
+++ b/api/common/charms/common.go
@@ -394,7 +394,7 @@ func convertCharmManifest(input *params.CharmManifest) (*charm.Manifest, error)
        }
        res := []charm.Base(nil)
        for _, v := range input.Bases {
-               str := fmt.Sprintf("%s/%s", v.Name, v.Channel)
+               str := fmt.Sprintf("%s@%s", v.Name, v.Channel)
                b, err := charm.ParseBase(str, v.Architectures...)
                if err != nil {
                        return nil, errors.Trace(err)
```

Deploy a charm `juju deploy ubuntu --base ubuntu@22.04`

### DIFF
--- a/base.go
+++ b/base.go
@@ -44,9 +44,12 @@ func (b Base) Validate() error {
 
 // String representation of the Base.
 func (b Base) String() string {
-	str := strings.ToLower(b.Name)
+	if b.Validate() != nil {
+		panic("cannot stringify invalid base. Bases should always be validated before stringifying")
+	}
+	str := b.Name
 	if !b.Channel.Empty() {
-		str += "/" + b.Channel.String()
+		str += "@" + b.Channel.String()
 	}
 	if len(b.Architectures) > 0 {
 		str = fmt.Sprintf("%s on %s", str, strings.Join(b.Architectures, ", "))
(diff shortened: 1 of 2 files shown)


base.go
```diff
@@ -46,28 +46,26 @@ func (b Base) Validate() error {
 func (b Base) String() string {
 	str := strings.ToLower(b.Name)
 	if !b.Channel.Empty() {
-		str += "/" + b.Channel.String()
+		str += "@" + b.Channel.String()
```
SimonRichardson, line 49: Turns out, this is going to be REALLY painful for the parsing logic. For example, if you have the following:

```go
base, err := ParseBase(Base{Name: "ubuntu"}.String())
// err == NotValid
```

base.go
```diff
@@ -46,28 +46,26 @@ func (b Base) Validate() error {
...
-	// Split the first forward-slash to get name and channel.
-	// E.g. "os/track/risk/branch" => ["os", "track/risk/branch"]
-	segments := strings.SplitN(s, "/", 2)
-	base.Name = strings.ToLower(segments[0])
-	channelName := ""
-	if len(segments) == 2 {
-		channelName = segments[1]
+	segments := strings.Split(s, "@")
+	if len(segments) != 2 {
+		return Base{}, errors.NotValidf("base string must contain exactly one @. %q", s)
```
SimonRichardson, line 65: So this isn't correct. You either need to say `@` is always mandatory and you fix `String()` to not return a base without an `@` symbol. Or you need to relax `ParseBase` and if it doesn't have an `@` symbol it just uses the name as the first segment.

Personally, you probably need to fix `String` and you may need to panic if there is a missing `channel`. I believe a panic __may__ be ok here, as `Validate` should ensure that a `channel` should be valid. An alternative may be to just always insert the `@` symbol. Although that will look weird. `ubuntu@` or `ubuntu@/risk/branch`

SimonRichardson (changes requested):

hmlanigan (commented):
I'll let Simon handle this change.

However we need to add another fix as we're breaking combat here already. Base defined Architecture: []string, however juju and charmhub have 1 Architecture per base: http://api.snapcraft.io/docs/charms.html

jack-w-shaw:
@hmlanigan unsure about that, since charm manifests have bases with a list of architectures
```
bases:
- architectures:
  - amd64
  channel: '16.04'
  name: ubuntu
- architectures:
  - amd64
  channel: '18.04'
  name: ubuntu
- architectures:
  - amd64
  channel: '20.04'
  name: ubuntu
```
Which at very least leaves room for multiple arches per base

PR #414, opened 2023-08-02
Add secret hooks

wallyworld:
The meta.Hooks() method was missing the secrets hooks. Add these.

Also update go.mod to address a CVE in `golang.org/x/net`.

### DIFF
--- a/hooks/hooks.go
+++ b/hooks/hooks.go
@@ -156,6 +156,13 @@ var secretHooks = []Kind{
 	SecretChanged, SecretExpired, SecretRemove, SecretRotate,
 }
 
+// SecretHooks returns all secret hook kinds.
+func SecretHooks() []Kind {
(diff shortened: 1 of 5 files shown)


hooks/hooks.go
```diff
@@ -18,7 +18,6 @@ const (
 	UpgradeCharm          Kind = "upgrade-charm"
 	Stop                  Kind = "stop"
 	Remove                Kind = "remove"
-	Action                Kind = "action"
```
hpidcock, line 21: this is used by the uniter

hpidcock: are you going to remove the code there?

hpidcock (approved):

hooks/hooks.go
wallyworld: ah, came up as unused for me when i searched, but i missed the usage, will add it back

PR #416, opened 2023-08-02
Add a new charm config option type secret;

ycliuhw:
Add a new charm config option type `secret`;

### DIFF
--- a/config.go
+++ b/config.go
@@ -55,11 +55,12 @@ var optionTypeCheckers = map[string]schema.Checker{
 	"int":     schema.Int(),
 	"float":   schema.Float(),
 	"boolean": schema.Bool(),
+	"secret":  schema.String(),
 }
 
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -55,11 +55,12 @@ var optionTypeCheckers = map[string]schema.Checker{
 	"int":     schema.Int(),
 	"float":   schema.Float(),
 	"boolean": schema.Bool(),
+	"secret":  schema.String(),
 }
 
 func (option Option) parse(name, str string) (val interface{}, err error) {
 	switch option.Type {
-	case "string":
+	case "string", "secret":
```
wallyworld, line 92: if it's a secret, we need to parse the URI to ensure it's valid. ie is the string a valid URI format and does it have scheme = secret and is the path component a valid XID and is the authority component (if present) a valid UUID

wallyworld (approved):
Looks good but we need to validate the secret URI format

PR #421, opened 2023-12-04
Add required metadata fields for rootless agents and containers.

hpidcock:
Adds new metadata fields required to enable rootless charms on Kubernetes.
- add the `uid` and `gid` fields to each container in the `containers` section which can be omitted.
- limit the uid and gid values to less than 1000 and greater than 9999 as per the spec.
- add the `charm-user` field which can be omitted, with values of `root`, `sudoer` and `non-root`.

### QA

Run unit tests.

### Links

Specs: [JU074](https://docs.google.com/document/d/1Y6OhDVWDadPYSlZVnPpH_upA_RNNRbM0luYLJJRrbUw/edit) [JU075](https://docs.google.com/document/d/1wSTB6R96B031j91ygK0hlfWdwaRPWCZ0ENQnjSDlEH4/edit)
Jira: JUJU-5123 JUJU-5127

### DIFF
--- a/base_test.go
+++ b/base_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/juju/utils/v3/arch"
 	gc "gopkg.in/check.v1"
 
-	"github.com/juju/charm/v11"
+	"github.com/juju/charm/v12"
 )
 
 type baseSuite struct {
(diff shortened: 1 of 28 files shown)


ycliuhw (approved):
LGTM, ty!

PR #424, opened 2024-02-13
Disallow unknow action parameters

Aflynn50:
Add a new default option `additionalProperties: false` to the json schema that is used when parsing actions. This option causes an error to be thrown by `ValidateParams` when the user passes a parameter that is not mentioned in the action description. 

This is a breaking change!!

To allow backwards compatibility or use actions in a prior way, use `additionalProperties: true`.

```yaml
echo:
  description: Provide an echo action.
  additionalProperties: true
  params:
    value:
      type: string
      description: The value that should be echo'd
```

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -127,10 +127,11 @@ func ReadActionsYaml(charmName string, r io.Reader) (*Actions, error) {
 		parallel := false
 		executionGroup := ""
 		thisActionSchema := map[string]interface{}{
-			"description": desc,
-			"type":        "object",
-			"title":       name,
-			"properties":  map[string]interface{}{},
+			"description":          desc,
+			"type":                 "object",
+			"title":                name,
+			"properties":           map[string]interface{}{},
+			"additionalProperties": false,
 		}
 
 		for key, value := range actionSpec {
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@ TAGS
 *.test
 *.sw[nop]
 
+.idea
 .vscode
 vendor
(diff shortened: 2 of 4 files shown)


SimonRichardson (commented):
You don't need to bump to v14 if we're landing this in main.

SimonRichardson (changes requested):
Just blocking this until we've removed the v14

actions.go
```diff
@@ -48,7 +48,8 @@ type ActionSpec struct {
 // ValidateParams validates the passed params map against the given ActionSpec
 // and returns any error encountered.
 // Usage:
-//   err := ch.Actions().ActionSpecs["snapshot"].ValidateParams(someMap)
+//
+//	err := ch.Actions().ActionSpecs["snapshot"].ValidateParams(someMap)
```
hmlanigan, line 52: remove me?

hmlanigan: Why is this a change? Looks like spacing or something

Aflynn50: Removed, I added all the linters from the CI to my Goland and have them set to rewrite the files, I should probs change that.

Aflynn50:
Removed the version bump @SimonRichardson

SimonRichardson (commented):
@Aflynn50 we just need a test where you have `additionalProperties: true` and it allows additional properties (we might already have one, just double-check). This should bake in this feature.

Aflynn50:
> @Aflynn50 we just need a test where you have `additionalProperties: true` and it allows additional properties (we might already have one, just double-check). This should bake in this feature.

Good point, added

SimonRichardson (approved):
This is a nice set of changes, with a nice light touch.

PR #425, opened 2024-02-21
Removes metrics code

SimonRichardson:
We no longer support metrics code, so we should remove the code. Moving it to dqlite would require understanding and maintaining code that has not full tested or integrated with romulus or omnibus is sometime. Fully understanding if all parts are correct would be challenging.

If in the future we want to reuse parts, we can always look back in the history of the repo to bring back the parts that are required (if any).


Aflynn50 (commented):
Checked out code and had a look for any remaining metrics/metering code, couldn't find any, looks good. I've approve but someone else with more context for this should review as well.

Aflynn50 (approved):

README.md
```diff
@@ -55,11 +55,6 @@ All the other fields are optional.
 The type can be either a `string`, `int`, `float` or `boolean`. Everything else
 will cause Juju to error out when reading the charm.
 
-### `metrics.yaml`
-
-`metrics.yaml` represents an optional metrics gathering configuration yaml. For
-more information about metrics, read up on [Metric collecting charms](https://discourse.juju.is/t/metric-collecting-charms/1125)
```
hmlanigan, line 61: We need to find where else in documentation this needs to be removed.

hmlanigan (approved):
LGTM

PR #427, opened 2024-05-09
Add methods to retrieve raw bundle string

jack-w-shaw:
Add this RawBundle method to Bundle interface and BundleDataSource

In Juju, we have come across a need to run some extra parsing + verification which we do not want to do in juju/charm. We want to keep the data model clean in juju/charm.

As such, we need to way to keep the raw bundle data around for later parsing

Add this to BundleDataSource, as this is what we use in Juju

However, we also implement our own BundleDataSource in Juju. built with the Bundle interface here. So we need to add this method to Bundle as well

You can see a draft implementation for this change here:
https://github.com/juju/juju/pull/17350

As a flyby, delete some incorrect comments

## QA Steps

Compile into juju form the following PR and run it's QA steps:
https://github.com/juju/juju/pull/17350

### DIFF
--- a/bundle.go
+++ b/bundle.go
@@ -15,6 +15,8 @@ import (
 type Bundle interface {
 	// Data returns the contents of the bundle's bundle.yaml file.
 	Data() *BundleData
+	// BundleBytes returns the raw bytes content of a bundle
+	BundleBytes() []byte
 	// ReadMe returns the contents of the bundle's README.md file.
 	ReadMe() string
 	// ContainsOverlays returns true if the bundle contains any overlays.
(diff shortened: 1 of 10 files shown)


bundle.go
```diff
@@ -15,6 +15,8 @@ import (
 type Bundle interface {
 	// Data returns the contents of the bundle's bundle.yaml file.
 	Data() *BundleData
+	// RawBundle returns the raw string content of a bundle
+	RawBundle() string
```
SimonRichardson: I'm struggling to like this name. Maybe `BundleBytes()`?

jack-w-shaw: Yeah that's fair

I'll obviously change the return type to `[]byte` as well

SimonRichardson (approved):

